PR #73, opened 2019-02-01
Auto Stitch UVs fixes and improvements

karl-:
This PR includes some minor organization, a bug fix for auto-stitch, and improvements to the `Convert to Auto` function in the UV Editor.

- Fixes https://unity3d.atlassian.net/browse/WB-1197
- Fixes https://unity3d.atlassian.net/browse/WB-1201
- Breaks the monolithic `UVEditing` class into partial classes
- Significantly improve the accuracy of "Convert to Auto UV" function
- Add ability for auto-stitched faces to remain auto unwrapped.


gabrielw-us:
Don't forget to add me as a reviewer if the change affects what the user will see or how they do things, please :) Even minor things, it's important.

Also it currently bugs out if you flip the normals of the object:

1. create pb cube
1. select a uv face, and swap it to manual
1. use auto-stitch to unwrap the entire cube, starting from that face
1. Flip the object normals
1. Select a face and do the auto-stitch action to an adjacent face- results in overlap between source and clicked face

gabrielw-us:
Good point from Karl- don't put bugs here, wait till merged 👍 
Will create separate Jira for the flip-normals-stitch issue, so I don't forget.

PR #82, opened 2019-02-25
Added dynamic picking distance based on current object hovered

antoine-unity:
[Jira: WB-996]

**Problem:** 
With the currently high picking distance being the only option, selecting another object near your current selection can be annoying (See image bellow).

selection

**Solution:**
A modifier is now applied to the distance of a vertex/edge if the user is currently hovering another probuilder mesh. This prioritizes the hovered mesh over the selected mesh.

I also added a slider for picking distance in preferences to make it available in developer mode. (Default value stays what the constant value used to be).

image

### DIFF
--- a/com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
+++ b/com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
@@ -23,7 +23,7 @@ namespace UnityEditor.ProBuilder
         public static Pref<float> pickingDistance = new Pref<float>("picking.pickingDistance", 128f, SettingsScope.User);
 
         [UserSetting]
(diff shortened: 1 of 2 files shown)


gabrielw-us:
Prefs: I'd be ok with removing the multiplier from prefs, but the basic distance is very useful.

Clicking on other object: sure, sounds good to me. I'll create a JIRA for that, and link it here...https://unity3d.atlassian.net/browse/WB-1257

gabrielw-us:
> Looks good, but I'd request we remove the multiplier pref (leave the main distance pref, though).

oops, I didn't pull correctly :D

antoine-unity:
> Looks good to me, though I'd still prefer that we not expose the `Picking Distance` setting. The current default feels good to me, and with the hovered object modifier I think we address the issues that would have made this necessary.

Picking distance is now part of Developer Mode

PR #85, opened 2019-02-25
Increase min width of UV inspector to disallow horizontal scrollbar

karl-:
Fixes WB-1261


gabrielw-us:
Looks good, but there's an odd bug when dragging the width down to minimum - some of the buttons get compressed, others not (see image). When mouse is released, the buttons snap back to correct.

image

gabrielw-us:
Additional: I can only resize from the bottom-right corner, not able to grab other corners, or the side, top, or bottom for resizing.

karl-:
@gabrielw-us fixed the button sizing, see 6ed58bf 

> Additional: I can only resize from the bottom-right corner, not able to grab other corners, or the side, top, or bottom for resizing.

I think that's out of the scope of this bug, but a good feature request. I opened a jira for it.

PR #87, opened 2019-02-25
fixed colliders and triggers debug display

neilunity:
rendering was disabled onEnterPlayMode, which was not valid if user was loading scenes from code.
Added logic to onSceneLoaded for the same effect. Also made sure renderer hideflags was turned off in builds.

### DIFF
--- a/com.unity.probuilder/Editor/EditorCore/EntityVisibility.cs
+++ b/com.unity.probuilder/Editor/EditorCore/EntityVisibility.cs
@@ -10,11 +10,6 @@ namespace UnityEditor.ProBuilder
     [InitializeOnLoad]
     internal static class EntityVisibility
     {
-        private static readonly Pref<bool> m_ShowDetail = new Pref<bool>("entity.detailVisible", true);
-        private static readonly Pref<bool> m_ShowMover = new Pref<bool>("entity.moverVisible", true);
-        private static readonly Pref<bool> m_ShowCollider = new Pref<bool>("entity.colliderVisible", true);
-        private static readonly Pref<bool> m_ShowTrigger = new Pref<bool>("entity.triggerVisible", true);
-
         static EntityVisibility()
         {
 #if UNITY_2017_2_OR_NEWER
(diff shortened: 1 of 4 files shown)


com.unity.probuilder/Editor/EditorCore/EntityVisibility.cs
```diff
@@ -1,31 +1,47 @@
...
     /// </summary>
     [InitializeOnLoad]
-    static class EntityVisibility
+    internal static class EntityVisibility
     {
-        static Pref<bool> m_ShowDetail = new Pref<bool>("entity.detailVisible", true);
-        static Pref<bool> m_ShowMover = new Pref<bool>("entity.moverVisible", true);
-        static Pref<bool> m_ShowCollider = new Pref<bool>("entity.colliderVisible", true);
-        static Pref<bool> m_ShowTrigger = new Pref<bool>("entity.triggerVisible", true);
+        private static readonly Pref<bool> m_ShowDetail = new Pref<bool>("entity.detailVisible", true);
```
karl-: This block of fields can be removed, they're no longer used

com.unity.probuilder/Editor/EditorCore/EntityVisibility.cs
```diff
@@ -48,40 +64,33 @@ public static void SetEntityVisibility(EntityType entityType, bool isVisible)
...
-            {
                 if (entityBehaviour.manageVisibility)
-                {
-                    // skip OnExit because OnEnter is operating on an instanced new scene, no changes will affect the actual scene
                     if (isEntering)
                         entityBehaviour.OnEnterPlayMode();
-                }
-            }
 
             if (!isEntering)
```
karl-: Everything from here down is also no longer used

com.unity.probuilder/Editor/EditorCore/EntityVisibility.cs
karl-: Ditto for `SetEntityVisibility(EntityType entityType, bool isVisible)`

PR #88, opened 2019-02-25
Fixed edge hover goes through selected mesh

antoine-unity:
[Jira: WB-1217]

**Problem:**
If the vertex is the closest point to the mouse, one of the 3 edges that connect that vertex will be hovered, even if it's behind the mesh.
image

**Solution:**
Check the distance to the camera when multiple edge have the closest distance to the mouse to determine which one is closer.

### DIFF
--- a/com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
+++ b/com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
@@ -29,6 +29,7 @@ namespace UnityEditor.ProBuilder
         static SceneSelection s_Selection = new SceneSelection();
         static List<VertexPickerEntry> s_NearestVertices = new List<VertexPickerEntry>();
         static List<GameObject> s_OverlappingGameObjects = new List<GameObject>();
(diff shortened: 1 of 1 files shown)


com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
```diff
@@ -533,6 +548,34 @@ static float EdgeRaycast(Vector3 mousePosition, ScenePickerPreferences pickerPre
             return Mathf.Infinity;
         }
 
+        static Edge GetClosestEdgeToCamera(Vector3[] positions, IEnumerable<Edge> edges)
+        {
+            var camPos = SceneView.lastActiveSceneView.camera.transform.position;
+            var closestDistToScreen = Mathf.Infinity;
+            Edge closest = default;
```
karl-: We still need to support legacy runtime for 2018.3, use `Edge.Empty`.

antoine-unity: Fixed d41affb

PR #91, opened 2019-02-26
Fixed Shape Tool create not being added to the undo stack

antoine-unity:
Using the shape tool to build an object should now dirty the scene and the action should be undoable.


antoine-unity:
[Jira: WB-1296]

PR #92, opened 2019-02-26
Add missing tool info in SceneView

karl-:
Prior to 4.0.0 there was a display for scene view tool deltas. It got lost in the tools refactor, and this PR adds it back in as a part of the `PositionTool`.


gabrielw-us:
Sidenote: seeing the total distance moved, as a separate value, would be a major benefit in non-global coords. Is there room to add that here?

PR #95, opened 2019-02-26
Fix Action window in UVEditor not scrollable.

thomas-tu:
**Goal of this PR:**
In the UV Editor, user can only scroll the UV view (grid) and not the actions window that is embedded. This PR fixes that issue by checking when we scroll if mouse is hovering the actions window. If so, we don't scroll the UV view and don't consume the event.

Fix this case: https://unity3d.atlassian.net/browse/WB-1260

### DIFF
--- a/com.unity.probuilder/Editor/EditorCore/UVEditor.cs
+++ b/com.unity.probuilder/Editor/EditorCore/UVEditor.cs
@@ -906,11 +906,13 @@ namespace UnityEditor.ProBuilder
                     break;
 
                 case EventType.ScrollWheel:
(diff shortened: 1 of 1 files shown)


gabrielw-us:
@thomas-tu Bug in branch: does not scroll if nothing selected, or if just one vertex selected.

karl-:
💯

gabrielw-us:
@thomas-tu did you see my note? is bugged :( I don't see any further commits after that?

gabrielw-us:
Created a new Jira since this got merged: https://unity3d.atlassian.net/browse/WB-1306

PR #97, opened 2019-02-27
Fixed deselecting vertices sometimes not working properly

antoine-unity:
[Jira: WB-1023]

When drag selecting, the _n_ shared vertices would get picked up instead of just one like with CTLR+left click (which is the correct behaviour). CTRL/Shift+left click now removes every vertex sharing the same shared vertex you clicked.

### DIFF
--- a/com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
+++ b/com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
@@ -29,6 +29,7 @@ namespace UnityEditor.ProBuilder
         static SceneSelection s_Selection = new SceneSelection();
         static List<VertexPickerEntry> s_NearestVertices = new List<VertexPickerEntry>();
         static List<GameObject> s_OverlappingGameObjects = new List<GameObject>();
+        static readonly List<int> s_IndexBuffer = new List<int>(16);
 
         public static ProBuilderMesh DoMouseClick(Event evt, SelectMode selectionMode, ScenePickerPreferences pickerPreferences)
         {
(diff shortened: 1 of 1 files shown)


com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
```diff
@@ -106,7 +107,21 @@ public static ProBuilderMesh DoMouseClick(Event evt, SelectMode selectionMode, S
...
+                        foreach (var vertex in sharedVertex)
+                        {
+                            var index = Array.IndexOf(mesh.selectedIndexesInternal, vertex);
+                            if (index < 0)
+                                continue;
+
+                            s_IndexBuffer.Add(index);
+                        }
+                        
+                        mesh.SetSelectedVertices(mesh.selectedIndexesInternal.SortedRemoveAt(s_IndexBuffer));
```
karl-: I don't think `selectedIndexesInternal` or `s_IndexBuffer` will be guaranteed to be ordered correctly.

com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
```diff
@@ -106,7 +107,21 @@ public static ProBuilderMesh DoMouseClick(Event evt, SelectMode selectionMode, S
...
+                        var sharedVertex = mesh.sharedVerticesInternal[sharedIndex];
+                        s_IndexBuffer.Clear();
+                        foreach (var vertex in sharedVertex)
+                        {
+                            var index = Array.IndexOf(mesh.selectedIndexesInternal, vertex);
+                            if (index < 0)
+                                continue;
+
+                            s_IndexBuffer.Add(index);
+                        }
```
karl-: This could be condensed to `mesh.GetCoincidentVertices(s_Selection.vertex, s_IndexBuffer);`

com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
antoine-unity: Now sorting indexes before passing them on to sorted remove at.

com.unity.probuilder/Editor/EditorCore/EditorSceneViewPicker.cs
antoine-unity: I need the index of the vertex in the selected indexes array and not the index of the vertex itself.

PR #101, opened 2019-02-28
Fix UV groups

antoine-unity:
[WB-1085]

**Issues Fixed:**
- Change to one member of a UV group would not replicate on all other member of the group
- The UV settings would be applied to every faces individually instead of being applied to all faces at once. For example, the face would all be rotated around their own pivot instead of being rotated around the pivot of the entire group.


com.unity.probuilder/Runtime/Core/ProBuilderMeshFunction.cs
```diff
@@ -403,6 +403,20 @@ public void RefreshUV(IEnumerable<Face> facesToRefresh)
                 mesh.SetUVs(3, m_Textures3);
         }
 
+        public void SetGroupUV(AutoUnwrapSettings settings, int group)
```
karl-: Private

PR #112, opened 2019-03-05
Expose combine objects action

karl-:
Makes `CombineMeshes.Combine` public, enabling merging of meshes from code.


karl-:
https://unity3d.atlassian.net/browse/WB-1356

PR #116, opened 2019-03-06
Save all settings on application quit

karl-:
Previously it was a requirement to call `ProBuillderSettings.Save` at some point after modifying a project setting. Now settings are automatically saved when the application exits.

https://unity3d.atlassian.net/browse/WB-1358

### DIFF
--- a/com.unity.probuilder/Content/Resources/Materials/FacePicker.mat
+++ b/com.unity.probuilder/Content/Resources/Materials/FacePicker.mat
@@ -21,4 +21,4 @@ Material:
     m_TexEnvs: []
     m_Floats: []
     m_Colors:
-    - _Tint: {r: 1, g: 1, b: 1, a: 1}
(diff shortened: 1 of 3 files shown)


com.unity.probuilder/Content/Resources/Materials/FacePicker.mat
```diff
@@ -21,4 +21,4 @@ Material:
     m_TexEnvs: []
     m_Floats: []
     m_Colors:
-    - _Tint: {r: 1, g: 1, b: 1, a: 1}
+    - _Tint: {r: 0, g: 0, b: 0, a: 1}
```
thomas-tu: Is this an intended change?

karl-: It changes based on what version of unity it's opened in, but it can't be ignored in git.

PR #118, opened 2019-03-11
Fix inconsistencies in handle and actions when `SelectMode` is changed

karl-:
Fixes the following issues related to `ProBuilderMesh` retaining distinct element selection caches.

- [x] Frame bounds focuses on element selection when in Object mode
- [x] Apply material incorrectly applies to face selection when in Object mode
- [x] Dimension overlay displays element bounds when in Object mode
- [x] Handle position when switching select modes does not update

### DIFF
--- a/com.unity.probuilder/Editor/EditorCore/DimensionsEditor.cs
+++ b/com.unity.probuilder/Editor/EditorCore/DimensionsEditor.cs
@@ -1,13 +1,23 @@
+#if UNITY_2019_1_OR_NEWER
+#define SHORTCUT_MANAGER
+#endif
+
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using UnityEditor.SettingsManagement;
+using UnityEditor.ShortcutManagement;
 using UnityEngine;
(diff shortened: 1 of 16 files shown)


com.unity.probuilder/Editor/EditorCore/DimensionsEditor.cs
```diff
@@ -46,10 +74,17 @@ void OnEnable()
 #else
             SceneView.onSceneGUIDelegate += OnSceneGUI;
 #endif
+            MeshSelection.objectSelectionChanged += OnObjectSelectionChanged;
+            ProBuilderMesh.elementSelectionChanged += OnElementSelectionChanged;
+            ProBuilderEditor.selectionUpdated += EditingMeshSelection;
```
thomas-tu: Shouldn't we keep the same convention as the two others callbacks? "On..." ?

karl-: 👍 
beef4a2

PR #119, opened 2019-03-11
Fix obj exporting separate faces

karl-:
This change enables the obj exporter to share vertex positions, normals, and textures. It reduces file sizes, as well as translates coincident vertices to other DCCs.

This also fixes an issue when attempting to copy a built-in texture resource.

### DIFF
--- a/com.unity.probuilder/Runtime/Core/IntVec2.cs
+++ b/com.unity.probuilder/Runtime/Core/IntVec2.cs
@@ -1,0 +1,75 @@
+using UnityEngine;
+
+namespace UnityEngine.ProBuilder
+{
+	/// <summary>
+	/// Vertex positions are sorted as integers to avoid floating point precision errors.
+	/// </summary>
+	struct IntVec2 : System.IEquatable<IntVec2>
+	{
+		public Vector2 vec;
+
+		public float x { get { return vec.x; } }
+		public float y { get { return vec.y; } }
+
+		public const float RESOLUTION = VectorHash.FLT_COMPARE_RESOLUTION;
+
+		public IntVec2(Vector2 vector)
+		{
+			this.vec = vector;
+		}
+
+		public override string ToString()
+		{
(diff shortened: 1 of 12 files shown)


com.unity.probuilder/Runtime/Core/IntVec4.cs
```diff
@@ -0,0 +1,81 @@
...
+	struct IntVec4 : System.IEquatable<IntVec4>
+	{
+		public Vector4 vec;
+
+		public float x { get { return vec.x; } }
+		public float y { get { return vec.y; } }
+		public float z { get { return vec.z; } }
+		public float w { get { return vec.w; } }
+
+		public const float k_Resolution = VectorHash.FLT_COMPARE_RESOLUTION;
```
antoine-unity: Wrong code style for public const

com.unity.probuilder/Runtime/Core/IntVec2.cs
```diff
@@ -0,0 +1,75 @@
...
+	/// Vertex positions are sorted as integers to avoid floating point precision errors.
+	/// </summary>
+	struct IntVec2 : System.IEquatable<IntVec2>
+	{
+		public Vector2 vec;
+
+		public float x { get { return vec.x; } }
+		public float y { get { return vec.y; } }
+
+		public const float RESOLUTION = VectorHash.FLT_COMPARE_RESOLUTION;
```
antoine-unity: Wrong code style for public const

com.unity.probuilder/Runtime/Core/IntVec4.cs
karl-: 16e8fdad

com.unity.probuilder/Runtime/Core/IntVec2.cs
karl-: 16e8fdad

PR #125, opened 2019-03-25
Fix Connect Edges throwing exception with degenerate tris

karl-:
Fixes fogbugz case 1077016

This also cleans up the functions to remove degenerate triangles and unused vertices, moving them to a new `MeshValidation` class. A new method is added (`MeshValidation.EnsureMeshIsValid`) which cleans up degenerate tris and unused vertices while keeping the selection intact.

### DIFF
--- a/com.unity.probuilder/Editor/EditorCore/RepairActions.cs
+++ b/com.unity.probuilder/Editor/EditorCore/RepairActions.cs
@@ -1,10 +1,7 @@
-using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using System.Text;
 using UnityEngine.ProBuilder;
 using UnityEngine.ProBuilder.MeshOperations;
-using UnityEditor;
 using UnityEngine;
 
 namespace UnityEditor.ProBuilder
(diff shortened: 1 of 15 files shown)


com.unity.probuilder/Runtime/MeshOperations/MeshValidation.cs
```diff
@@ -0,0 +1,306 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace UnityEngine.ProBuilder.MeshOperations
+{
+    /// <summary>
+    /// Methods for testing that mesh topology is correct.
```
thomas-tu: This class seems to do more than just testing ProBuilderMeshes with the Rebuild/Remove methods. Maybe add a few words more info about.

com.unity.probuilder/Editor/EditorCore/RepairActions.cs
```diff
@@ -139,17 +136,17 @@ public static void MenuRemoveDegenerateTriangles()
 
             foreach (ProBuilderMesh pb in InternalUtility.GetComponents<ProBuilderMesh>(Selection.transforms))
             {
-                pb.ToMesh();
-
-                int[] rm = pb.RemoveDegenerateTriangles();
-                count += rm != null ? rm.Length : 0;
+                var removed = MeshValidation.EnsureMeshIsValid(pb);
```
neilunity: returned value is not documented properly, change method signature or document correctly

karl-: 14789d9c

com.unity.probuilder/Runtime/MeshOperations/MeshValidation.cs
karl-: 14789d9c

PR #126, opened 2019-03-27
Add option to Export just mesh asset

karl-:
- splits the `Export Asset` option into two exporters, one for just the `Mesh` and one for `Mesh & Prefab` (latter is current behavior)
- adds an option to replace the original objects with the exported objects

Screen Shot 2019-03-27 at 3 52 45 PM

Screen Shot 2019-03-27 at 3 52 49 PM


com.unity.probuilder/Editor/MenuActions/Export/Export.cs
```diff
@@ -84,12 +85,23 @@ protected override void OnSettingsGUI()
 
             EditorGUI.BeginChangeCheck();
 
+            EditorGUI.BeginChangeCheck();
```
neilunity: one too many change checks ?

karl-: There's another `EndChangeCheck` near the bottom of this function

PR #134, opened 2019-04-14
Improved logic of Export Asset

phobos2077:
Related [forum thread](https://forum.unity.com/threads/export-asset-not-working-with-net-4-x-api.531936/).

What was done:
- When overwriting Mesh asset, correctly replace it without breaking existing references in GameObjects and Prefabs
- Use typed in name for Mesh asset (the same name as Prefab) when exporting Prefab (instead of generated internal mesh name)

What can still be improved:
- When replacing Prefabs, any prefab instances or prefab variants will be "reset" to the same state as the newly saved Prefab. Expected behavior is to correctly update existing prefab while retaining all overrides in Instances and Variants. Expected behavior can be achieved with `PrefabUtility.ReplacePrefab`, but this API is obsoleted for some reason.

See also [this gist](https://gist.github.com/phobos2077/5cc09cac2bc2cbe5c877d3cd11aac9db).


com.unity.probuilder/Editor/MenuActions/Export/ExportAsset.cs
```diff
@@ -217,13 +205,32 @@ static string ExportMesh(string path, ProBuilderMesh mesh)
             mesh.ToMesh();
             mesh.Refresh();
             mesh.Optimize();
-            mesh.mesh.name = name;
 
-            string meshPath = AssetDatabase.GenerateUniqueAssetPath(string.Format("{0}/{1}.asset", relativeDirectory, name));
+            var meshAsset = Object.Instantiate(mesh.mesh);
```
karl-: This breaks the `Replace Source` option. When exporting as an asset with `Replace Source` enabled, the `ProBuilderMesh` component should be removed from the `GameObject` and it's `MeshFilter.sharedMesh` should point to the saved asset.

com.unity.probuilder/Editor/MenuActions/Export/ExportAsset.cs
```diff
@@ -168,11 +157,11 @@ static string ExportPrefab(string path, ProBuilderMesh pb, bool replace)
...
 
-            string meshPath = AssetDatabase.GenerateUniqueAssetPath(string.Format("{0}/{1}.asset", relativeDirectory, pb.mesh.name));
+            string meshPath = string.Format("{0}/{1}.asset", relativeDirectory, name);
 
-            AssetDatabase.CreateAsset(pb.mesh, meshPath);
-
-            Mesh meshAsset = (Mesh)AssetDatabase.LoadAssetAtPath(meshPath, typeof(Mesh));
+            Mesh meshAsset = Object.Instantiate(pb.mesh);
+            meshAsset.name = name;
+            meshAsset = CreateOrReplaceAsset(meshAsset, meshPath);
```
karl-: If exporting a prefab with the same name as an existing mesh asset, the export throws errors and overwrites the mesh asset. It should generate a unique file name for the mesh asset since it did not prompt the user to overwrite the mesh.

karl-: It looks like exporting as a prefab (with any combination of settings) throws errors in 2019.1.0f1.

phobos2077:
Thanks, I'll review those issues in 2019.1.

com.unity.probuilder/Editor/MenuActions/Export/ExportAsset.cs
phobos2077: Testing now with 2019.1.0f2 - no errors on export as a Prefab. Tried overwriting existing prefab and creating a new one.

com.unity.probuilder/Editor/MenuActions/Export/ExportAsset.cs
phobos2077: Fixed

com.unity.probuilder/Editor/MenuActions/Export/ExportAsset.cs
phobos2077: @karl- about the mesh asset overwrite, maybe we should add another option to "Overwrite existing mesh"? This is a convenient behavior when I'm iterating and want to export and overwrite the whole prefab along with the asset, since they use the same base name. Having "Model.prefab", "Model.asset" and "Model 1.asset" will just confuse people.

phobos2077:
Also found and fixed what appears to be a vanilla bug - selecting multiple objects and attempting to export with "Replace Source" always results in a crash.

com.unity.probuilder/Editor/MenuActions/Export/ExportAsset.cs
karl-: > Edit: another solution is to show the second save file dialog for the asset, but only when it's about to overwrite. User can confirm overwrite or save at different path.

I prefer this approach over having an "Overwrite" toggle. A toggle is easily overlooked, and the mistake means irreversibly lost data. The downside to this of course being that now users see two popups when saving a prefab that will overwrite both the `.prefab` and `.asset`. However, I think we could skip showing the overwrite confirmation for the mesh in the case that the user has already approved a `.prefab` overwrite, and the corresponding mesh to be overwritten is associated with the prefab.

Ex,

### Scenario 1

1. Export mesh only to "Cube.asset"
2. Export mesh and prefab to "Cube.prefab" and "Cube.asset" 

*Result:* 

Prefab overwrite dialog, and Mesh overwrite dialog.

If prefab overwrite is canceled, cancel the operation. If prefab overwrite is confirmed but mesh is not, generate a unique asset name for the mesh and continue.

### Scenario 2

1. Export mesh and prefab to "Cube.prefab" and "Cube.asset"  
2. Export different mesh and prefab to "Cube.prefab" and "Cube.asset"

*Result:* 

Prefab overwrite dialog, mesh overwrite is implied.

phobos2077: @karl- what about batch export? There's only one dialog in this case - to select an output folder.

karl-: I think for now always generate a unique name on export.

phobos2077: @karl- Done. Should now work as your suggested.

PR #135, opened 2019-04-16
Restructure and Yamato config

karl-:
Restructures the repository such that the package is at root level, making it compatible with the yamato ci (and using git protocol in the manifest).

This also fixes a spattering of failing tests on Unity 2018.3, caused by floating point inaccuracies in the `Projection.VectorToProjectionAxis` method. I took the opportunity to simplify the test directory layout, and improve the naming of many of the tests.

@thomas-tu can you please review the Yamato configuration?
@neil can you please check my fix for `VectorToProjectionAxis`?


.yamato/upm-ci.yml
```diff
@@ -0,0 +1,99 @@
+test_editors:
+  - version: 2018.3
+  - version: 2019.1
+  - version: trunk
```
thomas-tu: As trunk is now 2019.3. We should add 2019.2 here.

PR #139, opened 2019-05-03
Ensure that the SelectVertexColor handles selecting faces without col…

MartinImgSpc:
…or case.

Add tests to cover this SelectVertexColor action 2 main cases: selecting a face with color and without color.

### DIFF
--- a/Tests/Editor/Selection/SelectFacesByColor.cs
+++ b/Tests/Editor/Selection/SelectFacesByColor.cs
@@ -1,0 +1,140 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+using UObject = UnityEngine.Object;
+using NUnit.Framework;
+using UnityEngine.ProBuilder;
+using UnityEditor.ProBuilder;
+using UnityEditor;
+using UnityEngine.TestTools;
+
+
+namespace UnityEngine.ProBuilder.EditorTests.Selection
+{
+    class SelectFacesByColor
+    {
+        ProBuilderMesh[] selectables;
+        List<Color> faceColors;
+
+        void Setup(bool withColor = false)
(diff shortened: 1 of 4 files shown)


Tests/Editor/Selection/SelectFacesByColor.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+            selectables = new ProBuilderMesh[]
+            {
+                shape1,
+                shape2
+            };
+        }
+
+        void Cleanup()
```
vkovec: If you add [TearDown] attribute this will automatically happen after each test, whether or not the test fails.

Tests/Editor/Selection/SelectFacesByColor.cs
```diff
@@ -0,0 +1,140 @@
...
+
+
+namespace UnityEngine.ProBuilder.EditorTests.Selection
+{
+    class SelectFacesByColor
+    {
+        ProBuilderMesh[] selectables;
+        List<Color> faceColors;
+
+        void Setup(bool withColor = false)
```
vkovec: I'm not sure if this would work since you have an argument, but adding [SetUp] attribute would make it so this gets called automatically before each test

Tests/Editor/Selection/SelectFacesByColor.cs
MartinImgSpc: Makes sense! Thank you!

Tests/Editor/Selection/SelectFacesByColor.cs
MartinImgSpc: Given the argument in the setup it will not work this time. But I will know for the future! Thank you!

PR #140, opened 2019-05-03
Allow picking on a Terrain when trying to get the best plane.

MartinImgSpc:
Modified FindBestPlane to handle picking Terrain.
Note: Tried to create a unit test without success because FindBestPlane relies on UnityEditor.HandleUtility.PickGameObject which cannot be called outside OnGui calls.

### DIFF
--- a/Editor/EditorCore/EditorHandleUtility.cs
+++ b/Editor/EditorCore/EditorHandleUtility.cs
@@ -512,16 +512,13 @@ namespace UnityEditor.ProBuilder
                     ignorePicking.Add(go);
 
                 go = HandleUtility.PickGameObject(mousePosition, false, ignorePicking.ToArray());
-            } while (go != null && go.GetComponent<MeshFilter>() == null);
+            } while (go != null && (go.GetComponent<MeshFilter>() == null && go.GetComponent<Terrain>() == null));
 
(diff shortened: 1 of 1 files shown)


thomas-tu:
The current changes would help pick the first point on a terrain and define a plane that would be used to build the polyshape but won't prevent creating points behind a terrain on that plane. Is it something we want for the Polyshape tool? @karl-  @gabrielw-us 

I would maybe expect to have my first three points stick to terrain (3 points to build a plane vs. using the first point normal).

gabrielw-us:
> I would maybe expect to have my first three points stick to terrain (3 points to build a plane vs. using the first point normal).

That could be useful as a separate feature, sure! I'd say make a separate feature JIRA for that, we can discuss? For this task, I just wanted to ensure the standard functionality works (first click sets initial position and plane).

karl-:
> I would maybe expect to have my first three points stick to terrain (3 points to build a plane vs. using the first point normal).

I think this would probably be good to support for any object intersecting the first 3 rays. However I agree with @gabrielw-us, let's create a separate case for that.

thomas-tu:
Thanks guys!

MartinImgSpc:
Thank you!

PR #148, opened 2019-05-10
Fix obj export with multiple meshes creating invalid files

karl-:
WB-1492

Fixes an issue where the obj exporter would incorrectly offset additional mesh groups by the original vertex count instead of the coincident merged counts.

Adds new unit tests for obj export.


gabrielw-us:
@karl- Can I get a link to the JIRA issue? Not understanding what I should be checking here. Thanks!

thomas-tu:
@gabrielw-us https://unity3d.atlassian.net/browse/WB-1492

PR #150, opened 2019-05-10
Change default value of "Apply Transforms" in Export options

karl-:
WB-1491

Default for exporting individual meshes is now `Apply Transforms: False`. Transforms are still always applied when exporting as a combined mesh.


karl-:
> Need to also disable "Export as Group" by default. The idea is to export the thing(s) without changing them, by default - so if I have 12 object selected, I get 12 OBJs, and all 12 keep their original transforms, meshes, etc.

This I'm less inclined to change - it seems to me that "Export" would usually imply a single result, and that our batch version of export is a variant. 

It also makes it much easier to export a scene with many objects and have it import to another DCC correctly. The OBJ spec does not have a concept of transforms afaik, so this makes round-tripping possible.

PR #152, opened 2019-05-10
Collect coincident vertices when doing split vertices action

karl-:
Fixes WB-1359

Split Vertex action now operates on all coincident vertices in the selection, not just the selected indices.


karl-:
@gabrielw-us  fixed [these errors](https://github.com/Unity-Technologies/com.unity.probuilder/pull/152#pullrequestreview-237376156), but with known issue:

```
                // This isn't completely accurate, because to check that each selected vertex has coincident vertices is
                // unnecessarily expensive for the purposes of this property. So here we handle the most common case,
                // where a single vertex is selected with no additional coincident vertices.
                return base.enabled
                    && MeshSelection.selectedVertexCountObjectMax > 0
                    && !(MeshSelection.selectedVertexCountObjectMax == 1 && MeshSelection.selectedCoincidentVertexCountMax == 1);
```

PR #154, opened 2019-05-10
Wb-1297 fix cone UVs

vkovec:


Runtime/MeshOperations/UV/UvUnwrapping.cs
```diff
@@ -7,12 +7,30 @@ static class UvUnwrapping
...
+            Projection.PlanarProject(mesh, face, projection != Vector3.zero? projection : Vector3.zero);
             ApplyUVSettings(mesh.texturesInternal, face.distinctIndexesInternal, face.uv);
         }
 
+        /// <summary>
+        /// Copy UVs from faceToCopy to face for the given mesh.
+        /// </summary>
+        /// <param name="mesh">ProbuilderMesh</param>
+        /// <param name="face">face to copy UVs to</param>
+        /// <param name="faceToCopy">face to copy UVs from</param>
```
MartinImgSpc: Detail but description and name is a bit confusing to me at first. Also I would kind of of expect the argument to be the other way around source -> destination

PR #155, opened 2019-05-10
add 'off' to dimensions overlay toggle cycles

karl-:
WB-1367


karl-:
> Also - probably best to not show dimensions if in Object mode, and using Element Dimensions?

The preference it's toggling is `Always use Object Bounds: [Yes, No]`. I'll change the behaviour and make the preference an enum `Bounds Display: [ Object, Element ]`

PR #156, opened 2019-05-17
Add an offset elements menu action

karl-:
Replacing the "Offset" button and vector field in the Inspector, a new toolbar action "Move Elements" is introduced.

It has two fields to set, `Space: [ World, Local ]` and `Move: XYZ`.

image

### DIFF
--- a/Editor/EditorCore/EditorToolbarLoader.cs
+++ b/Editor/EditorCore/EditorToolbarLoader.cs
@@ -85,6 +85,7 @@ namespace UnityEditor.ProBuilder
 
                 // All
                 new Actions.SetPivotToSelection(),
+                new Actions.MoveElements(),
 
                 // Faces (All)
                 new Actions.DeleteFaces(),
--- a/Editor/EditorCore/EditorToolbarMenuItems.cs
+++ b/Editor/EditorCore/EditorToolbarMenuItems.cs
@@ -415,6 +415,21 @@ namespace UnityEditor.ProBuilder
 				EditorUtility.ShowNotification(instance.DoAction().notification);
 		}
 
+		[MenuItem(k_MenuPrefix + "Geometry/Move Elements ", true)]
+		static bool MenuVerify_MoveElements()
+		{
+			var instance = EditorToolbarLoader.GetInstance<MoveElements>();
+			return instance != null && instance.enabled;
+		}
+
+		[MenuItem(k_MenuPrefix + "Geometry/Move Elements", false, PreferenceKeys.menuGeometry + 3)]
+		static void MenuPerform_MoveElements()
+		{
+			var instance = EditorToolbarLoader.GetInstance<MoveElements>();
+			if(instance != null && instance.enabled)
+				EditorUtility.ShowNotification(instance.DoAction().notification);
+		}
+
 		[MenuItem(k_MenuPrefix + "Geometry/Set Pivot To Selection ", true)]
 		static bool MenuVerify_SetPivotToSelection()
 		{
--- a/Editor/MenuActions/Geometry/MoveElements.cs
+++ b/Editor/MenuActions/Geometry/MoveElements.cs
@@ -1,0 +1,101 @@
+using UnityEngine;
+using UnityEngine.ProBuilder;
+
+namespace UnityEditor.ProBuilder.Actions
+{
+    sealed class MoveElements : MenuAction
+    {
+        enum CoordinateSpace
+        {
+            Local,
+            World
+        }
+
+        static readonly TooltipContent s_TooltipFace = new TooltipContent ( "Move Faces", "Move the selected elements by a set amount." );
+        static readonly TooltipContent s_TooltipEdge = new TooltipContent ( "Move Edges", "Move the selected elements by a set amount." );
+        static readonly TooltipContent s_TooltipVert = new TooltipContent ( "Move Vertices", "Move the selected elements by a set amount." );
+
+        static Pref<Vector3> s_MoveDistance = new Pref<Vector3>("MoveElements.s_MoveDistance", Vector3.up);
+        static Pref<CoordinateSpace> s_CoordinateSpace = new Pref<CoordinateSpace>("MoveElements.s_CoordinateSpace", CoordinateSpace.World);
+
+        public override ToolbarGroup group { get { return ToolbarGroup.Geometry; } }
+        public override Texture2D icon { get { return IconUtility.GetIcon(null, IconSkin.Pro); } }
+
+        public override TooltipContent tooltip
+        {
+            get
+            {
+                if(ProBuilderEditor.selectMode == SelectMode.Face)
+                    return s_TooltipFace;
+                if(ProBuilderEditor.selectMode == SelectMode.Edge)
+                    return s_TooltipEdge;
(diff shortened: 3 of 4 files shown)


gabrielw-us:
@karl- in this branch, as soon as I enable any element mode I get errors and the PB toolbar goes blank :(

ArgumentNullException: Value cannot be null.

Parameter name: key
System.Collections.Generic.Dictionary`2[TKey,TValue].FindEntry (TKey key) (at <1f0c1ef1ad524c38bbc5536809c46b48>:0)
System.Collections.Generic.Dictionary`2[TKey,TValue].TryGetValue (TKey key, TValue& value) (at <1f0c1ef1ad524c38bbc5536809c46b48>:0)
UnityEditor.ProBuilder.IconUtility.GetIcon (System.String iconName, UnityEditor.ProBuilder.IconSkin skin) (at C:/_git/com.unity.probuilder/Editor/EditorCore/IconUtility.cs:45)
UnityEditor.ProBuilder.Actions.MoveElements.get_icon () (at C:/_git/com.unity.probuilder/Editor/MenuActions/Geometry/MoveElements.cs:22)
UnityEditor.ProBuilder.EditorToolbar.IsActionValid (UnityEditor.ProBuilder.MenuAction action) (at C:/_git/com.unity.probuilder/Editor/EditorCore/EditorToolbar.cs:217)
UnityEditor.ProBuilder.EditorToolbar.OnGUI () (at C:/_git/com.unity.probuilder/Editor/EditorCore/EditorToolbar.cs:230)
UnityEditor.ProBuilder.ProBuilderEditor.OnGUI () (at C:/_git/com.unity.probuilder/Editor/EditorCore/ProBuilderEditor.cs:487)
System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) (at <1f0c1ef1ad524c38bbc5536809c46b48>:0)
Rethrow as TargetInvocationException: Exception has been thrown by the target of an invocation.
System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) (at <1f0c1ef1ad524c38bbc5536809c46b48>:0)
System.Reflection.MethodBase.Invoke (System.Object obj, System.Object[] parameters) (at <1f0c1ef1ad524c38bbc5536809c46b48>:0)
UnityEditor.HostView.Invoke (System.String methodName, System.Object obj) (at C:/buildslave/unity/build/Editor/Mono/HostView.cs:345)
UnityEditor.HostView.Invoke (System.String methodName) (at C:/buildslave/unity/build/Editor/Mono/HostView.cs:339)
UnityEditor.HostView.InvokeOnGUI (UnityEngine.Rect onGUIPosition, UnityEngine.Rect viewRect) (at C:/buildslave/unity/build/Editor/Mono/HostView.cs:315)
UnityEditor.DockArea.DrawView (UnityEngine.Rect viewRect, UnityEngine.Rect dockAreaRect, System.Boolean floatingWindow, System.Boolean isBottomTab) (at C:/buildslave/unity/build/Editor/Mono/GUI/DockArea.cs:373)
UnityEditor.DockArea.OldOnGUI () (at C:/buildslave/unity/build/Editor/Mono/GUI/DockArea.cs:340)
UnityEngine.UIElements.IMGUIContainer.DoOnGUI (UnityEngine.Event evt, UnityEngine.Matrix4x4 parentTransform, UnityEngine.Rect clippingRect, System.Boolean isComputingLayout, UnityEngine.Rect layoutSize) (at C:/buildslave/unity/build/Modules/UIElements/IMGUIContainer.cs:278)
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr)

gabrielw-us:
Update: issue is that an icon is needed in icon mode. Per discussion, that should "fail better" though.

Also, action notification should use term that matches element mode, ie "Move 1 Face" vs "Move 4 elements" in Face mode.

gabrielw-us:
Icon in progress

gabrielw-us:
Ok, added icon, but I want to think on a better method of choosing/showing Space. Hold on ...

gabrielw-us:
Okay, 1 change: this should not have a space selection, instead use the currently active ProBuilder Coordinate Space (World, Local, Planar). This will make the action 100% predictable (it will follow the Handles).

karl-:
> Okay, 1 change: this should not have a space selection, instead use the currently active ProBuilder Coordinate Space (World, Local, Planar). This will make the action 100% predictable (it will follow the Handles).

I really like the idea of respecting existing preferences. Unfortunately in this case these aren't exactly the same thing. ProBuilder's `Orientation: ` toggle only affects the rotation of the handle, not the working coordinate space.

If we follow the handle behavior, that is removing the functionality of the `Space: World, Local` as it works currently.  So for example, if you wanted to move elements along the `X` axis in model space, that is no longer possible. The movement would always take place in world space, following the `X` direction of the handle.

gabrielw-us:
I don't see the issue, for example - in the image below, I can set the Handle orientation, then use "Move Elements" with nice predictable results because the Handle tells me exactly where it's going to move, yes?

image

gabrielw-us:
Shoot, where was I on this ... sorry, got all kerfluffled and forgot kept putting it off ... now we are getting yelled at by forum folk ... pretty sure I had realized, after our chat, that it made more sense to do it slightly different ... 30 min

gabrielw-us:
Okay, several upgrades I would like to see:
- Let user choose "Handle" so they have visual guide, and most importantly, can change the direction without opening settings, ie they just toggle PB orientation mode, much more fluid
- Let user break from Handle mode if they prefer, of course
- Let user choose if elements follow the active, or all move on their own (especially useful for expanding the radius of an item, or randomizing)
- Let user choose if Scale affects the actual move distance (looks like this is always on right now, and makes sense as a default, but I may want my elements to move by exact amounts, regardless of scale)

An example:

image

gabrielw-us:
... iff'n a feller really wanted to get interesting, they'd add some buttons to quick-move on each axis too!

image

karl-:
discussion today with @gabrielw-us 

## todo

- [ ] `Coordinate Space: { World, Local, Element, Handle }`*
- [ ] No "follow active" or "scale" toggles (already covered by `Coordinate Space`)
- [ ] Add the ability to persist menu option windows
- [ ] Show a line + dot preview of each selected element group's offset

|--|--|
| World | Offset value is specified in world space, vertex transformations are applied in world space |
| Local | Offset value is specified in active object model space, vertex transformations are applied in model space for each object |
| Element | Offset value is specified in active element space, vertex transformations are applied in element space for each element group. |
| Handle | Offset value is specified in handle space, vertex transformations are applied in world space |

karl-:
image

@gabrielw-us here's what the current preview looks like. the color is tied to the direction of the offset

gabrielw-us:
Note so I don't forget: lines are not showing up on my windows machine, lots of "matrix depth stack" error spam. 19.2, 19.3 repro.

gabrielw-us:
Edit: no more error spam in the latest Git, but line is still not visible (19.2)

PR #158, opened 2019-05-24
WB-1106 Apply material to selected faces only if Edit UVs in Scene mode enabled

vkovec:
- previously was applying material to the entire mesh even if there were faces selected


Editor/EditorCore/SceneDragAndDropListener.cs
```diff
@@ -47,7 +47,7 @@ public static bool isDragging
 
         static bool isFaceMode
         {
-            get { return ProBuilderEditor.selectMode == SelectMode.Face; }
+            get { return ProBuilderEditor.selectMode == SelectMode.Face || ProBuilderEditor.selectMode == SelectMode.TextureFace; }
```
vkovec: @MartinImgSpc noticed that in TextureTool.cs, isFaceMode checks if selectMode contains flags instead of checking if it matches a single flag like here:
```
var isFaceMode = ProBuilderEditor.selectMode.ContainsFlag(SelectMode.TextureFace | SelectMode.Face);
```
Would this be preferable? Let me know, thanks!

karl-: Either one is fine, though personally I prefer the `ContainsFlag` version.

PR #159, opened 2019-05-28
Ensure that when in append mode and re-clicking on a Face/Edge/Vertex…

MartinImgSpc:
… which is already selected will only change the active element, unless you are clicking again on the active element which would lead to unselection.


Editor/EditorCore/EditorSceneViewPicker.cs
```diff
@@ -83,7 +83,17 @@ public static ProBuilderMesh DoMouseClick(Event evt, SelectMode selectionMode, S
...
+                        if (!appendModifier || s_Selection.face == mesh.GetActiveFace())
+                        {
+                            mesh.RemoveFromFaceSelectionAtIndex(sel);
+                        }
+                        else
+                        {
+                            mesh.selectedFaceIndicesInternal = mesh.selectedFaceIndicesInternal.Remove(ind);
+                            mesh.SetSelectedFaces(mesh.selectedFaceIndicesInternal.Add(ind));
+                        }
+                    }
```
karl-: If `SelectionModifierBehaviour` now changes the behavior of mouse clicking, we should support all cases. ex,

```
                    if (sel > -1)
                    {
                        if (pickerPreferences.selectionModifierBehavior == SelectionModifierBehavior.Add)
                        {
                            // Make the selected face 'active' by making it the last in the selection array
                            mesh.RemoveFromFaceSelectionAtIndex(sel);
                            mesh.AddToFaceSelection(ind);
                        }
                        else
                        {
                            // If the face is in selection, appendModifier is true. In this case Subtract and Difference
                            // provide the same result.
                            mesh.RemoveFromFaceSelectionAtIndex(sel);
                        }
                    }
                    else
                    {
                        mesh.AddToFaceSelection(ind);
                    }```

karl-: Note - it looks like we aren't treating mouse click vertex selection as shared vertices. We  should be.

gabrielw-us:
You have no idea how happy this makes me :D Checking ASAP!

gabrielw-us:
Looking good! 2 problems though:
1) Active Element is wrong if using multiple shapes
- Create two pb cubes
- Select one, enter vert mode
- Click a vertex
- SHIFT-click a vert on the other cube (it becomes Active, correct)
- SHIFT-click a vert on the first cube (it does not become Active, incorrect)

2) CTRL-click does not remove from selection
- create several shapes
- ctrl-click, note that selection is added or removed
- do same with pb elements, note that selection is always added (incorrect)

Editor/EditorCore/EditorSceneViewPicker.cs
MartinImgSpc: The selection modifier behavior is for drag selection and is documented as such. Do we want to change the picking behavior in the same way? As per @gabrielw-us he would like the behavior of the ctrl+click to be substract and shift to be add. Doesn't this conflict with using the preference for the shift?

MartinImgSpc:
> Looking good! 2 problems though:
> 
> 1. Active Element is wrong if using multiple shapes
> 
> * Create two pb cubes
> * Select one, enter vert mode
> * Click a vertex
> * SHIFT-click a vert on the other cube (it becomes Active, correct)
> * SHIFT-click a vert on the first cube (it does not become Active, incorrect)
> 
> 1. CTRL-click does not remove from selection
> 
> * create several shapes
> * ctrl-click, note that selection is added or removed
> * do same with pb elements, note that selection is always added (incorrect)

Good catches, will address today.

gabrielw-us:
Thanks!!

gabrielw-us:
Problem: Active Element is set incorrectly if CTRL-click to deselect on non-active object

1. two pb cubes
1. select 3 verts on one cube - remember the order!
1. SHIFT or CTRL to add vert from the other cube (it becomes Active Object)
1. CTRL-click to remove one of the first cube's 3 selected verts (on the Click, it is not Active Object)

Expected: Active element does not change
Result: Active element changes to the other cube's last selected vert, Handle jumps there

karl-:
Currently `Shift Click` always either adds to the selection, or changes the active selection. When selecting GameObjects in the SceneView, `Shift Click` will set the the active selection if the selected object is in the selection but not active, and if it is the active selection, it removes it from the selection. I'd expect we want to match the scene behaviour, @gabrielw-us was this your intent?

PR #160, opened 2019-05-28
WB-1044 drag-selecting single vertex enables collapse verts

vkovec:


package.json
```diff
@@ -1,7 +1,7 @@
 {
 				"name": "com.unity.probuilder",
 				"displayName": "ProBuilder",
-				"version": "4.1.0-preview.5",
+				"version": "4.1.0-preview.6",
```
karl-: This doesn't match the changelog. I think we're on `4.1.0-preview.5` currently.

PR #161, opened 2019-05-30
Allow resizing UVEditor action window from any corner,sides.

MartinImgSpc:


Editor/EditorCore/EditorGUILayout.cs
```diff
@@ -120,32 +121,118 @@ class ResizeHandleState
             public Rect startingRect;
         }
 
-        public static Rect DoResizeHandle(Rect rect)
+        private delegate Rect ComputeResize(Rect curretRect, ResizeHandleState handleState, Event currentEvent);
```
karl-: style guide - omit implicit access modifiers

karl-: `Rect curretRect` <- `currentRect`

Editor/EditorCore/EditorGUILayout.cs
```diff
@@ -120,32 +121,118 @@ class ResizeHandleState
...
+            return currentRect;
+        }
+
+        static int s_ResizeHandleAreaDimension = 6;
+        static int s_MoveWindowAreaHeight = 30;
+
+        public static Rect DoResizeHandle(Rect rect)
+        {
+            var evt = Event.current;
+            if (evt.rawType == UnityEngine.EventType.Used)
```
karl-: Why use `rawType` here? I would expect you'd want the `type`.

MartinImgSpc:
@karl- and other reviewers. I will do the change proposed. However when we have done supplemental testing we discovered a behavior which is incorrect. When the Action window reaches its minimal width or height and manipulating from left or top, it would end up swimming rather than sticking into position. I will do the changes to address this and update this review with the changes requested. Sorry for the inconvenience.

thomas-tu:
> Content looks good to me. Approving with a request that redundant `private` modifiers be removed prior to merge.
> 
> Edit - Also please change `evt.rawType` -> `evt.type`.

+1 with the requested changes above ^

PR #163, opened 2019-06-06
Wb 152 allow zero height polyshape

MartinImgSpc:
If there is extrusion the other faces will be created otherwise it will remain single sided.

### DIFF
--- a/Editor/EditorCore/PolyShapeEditor.cs
+++ b/Editor/EditorCore/PolyShapeEditor.cs
@@ -609,7 +609,8 @@ namespace UnityEditor.ProBuilder
 
                 case KeyCode.Escape:
                 {
-                    if (polygon.polyEditMode == PolyShape.PolyEditMode.Path)
+                    if (polygon.polyEditMode == PolyShape.PolyEditMode.Path ||
+                        polygon.polyEditMode == PolyShape.PolyEditMode.Height)
                     {
                         DiscardIncompleteShape();
                         evt.Use();
@@ -619,12 +620,7 @@ namespace UnityEditor.ProBuilder
                         SetPolyEditMode(PolyShape.PolyEditMode.None);
                         evt.Use();
                     }
-                    else if (polygon.polyEditMode == PolyShape.PolyEditMode.Height)
-                    {
-                        SetPolyEditMode(PolyShape.PolyEditMode.None);
-                        evt.Use();
-                    }
-
+                    
                     break;
                 }
             }
--- a/Editor/EditorCore/PolyShapeEditor.cs
+++ b/Editor/EditorCore/PolyShapeEditor.cs
@@ -609,13 +609,17 @@ namespace UnityEditor.ProBuilder
 
                 case KeyCode.Escape:
                 {
-                    if (polygon.polyEditMode == PolyShape.PolyEditMode.Path ||
-                        polygon.polyEditMode == PolyShape.PolyEditMode.Height)
+                    if (polygon.polyEditMode == PolyShape.PolyEditMode.Path)
                     {
                         DiscardIncompleteShape();
                         evt.Use();
                     }
                     else if (polygon.polyEditMode == PolyShape.PolyEditMode.Edit)
+                    {
+                        SetPolyEditMode(PolyShape.PolyEditMode.None);
+                        evt.Use();
+                    }
+                    else if (polygon.polyEditMode == PolyShape.PolyEditMode.Height)
                     {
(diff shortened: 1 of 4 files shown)


Runtime/MeshOperations/AppendElements.cs
```diff
@@ -264,19 +264,23 @@ public static ActionResult CreateShapeFromPolygon(this ProBuilderMesh mesh, ILis
                 mesh.sharedVerticesInternal = SharedVertex.GetSharedVerticesWithPositions(vertices);
                 mesh.InvalidateCaches();
 
-                Vector3 nrm = Math.Normal(mesh, mesh.facesInternal[0]);
+                
+                if (extrude != 0.0f )
```
vkovec: is it easy to get it to be 0.0f exactly?

Tests/Runtime/MeshOps/AppendElementsTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        [Test]
+        public void CreateShapeFromPolygon_ExtrudeOrNot()
+        {
+            m_Poly.m_Points.Add(new Vector3(0,0,0));
+            m_Poly.m_Points.Add(new Vector3(0,0,2));
+            m_Poly.m_Points.Add(new Vector3(2,0,0));
+            m_Poly.extrude = 0.0f;
+            m_Poly.flipNormals = false;
```
vkovec: what happens if flipNormals is true? is it possible for it to be?

Runtime/MeshOperations/AppendElements.cs
MartinImgSpc: Yes it is the constructor initialization of the member of the PolyShape that is passed.

Tests/Runtime/MeshOps/AppendElementsTests.cs
MartinImgSpc: It is just used when creating the opposite face. actually if I do not initialize it is the default value false .

gabrielw-us:
This look great, super useful! Only 1 issue, I noticed that planes are created facing away from user if they are created into "space" (vs onto a surface).

Editor/EditorCore/PolyShapeEditor.cs
```diff
@@ -213,7 +213,7 @@ void RebuildPolyShapeMesh(bool vertexCountChanged = false)
             DrawPolyLine(polygon.m_Points);
 
             if(polygon.polyEditMode != PolyShape.PolyEditMode.Path)
-                polygon.CreateShapeFromPolygon();
```
vkovec: is this function still used?

Runtime/MeshOperations/AppendElements.cs
```diff
@@ -229,6 +229,20 @@ internal static ActionResult CreateShapeFromPolygon(this PolyShape poly)
         /// <param name="flipNormals">If true the faces will be inverted at creation.</param>
         /// <returns>An ActionResult with the status of the operation.</returns>
         public static ActionResult CreateShapeFromPolygon(this ProBuilderMesh mesh, IList<Vector3> points, float extrude, bool flipNormals)
```
vkovec: this is maybe a style preference, but why not make Vector3.up the default value for cameraLookAt?

Editor/EditorCore/PolyShapeEditor.cs
```diff
@@ -213,7 +213,7 @@ void RebuildPolyShapeMesh(bool vertexCountChanged = false)
...
+                polygon.mesh.CreateShapeFromPolygon(polygon.m_Points, polygon.extrude, polygon.flipNormals, SceneView.lastActiveSceneView.camera.transform.forward);
```
vkovec: just a suggestion but instead of passing all these here, you could modify CreateShapeFromPolygon() to take the cameraLookAt as an argument

Runtime/MeshOperations/AppendElements.cs
MartinImgSpc: Given that it is a public API it needs to be unchanged as far as I understood.

Editor/EditorCore/PolyShapeEditor.cs
MartinImgSpc: With your suggestion it will be again.

Editor/EditorCore/PolyShapeEditor.cs
MartinImgSpc: Indeed more elegant.

PR #164, opened 2019-06-07
WB-112 invalidate bad polyshapes

vkovec:


Runtime/MeshOperations/AppendElements.cs
```diff
@@ -250,9 +260,18 @@ public static ActionResult CreateShapeFromPolygon(this ProBuilderMesh mesh, ILis
             {
                 int[] indexes = triangles.ToArray();
 
+                // check that all points are represented in the triangulation
+                HashSet<int> triangleSet = new HashSet<int>(triangles);
```
MartinImgSpc: Would it make sense to have a unit test to check this condition?

Editor/EditorCore/PolyShapeEditor.cs
```diff
@@ -212,8 +218,31 @@ void RebuildPolyShapeMesh(bool vertexCountChanged = false)
...
             DrawPolyLine(polygon.m_Points);
 
-            if(polygon.polyEditMode != PolyShape.PolyEditMode.Path)
-                polygon.CreateShapeFromPolygon();
+            if (polygon.polyEditMode != PolyShape.PolyEditMode.Path)
+            {
+                var result = polygon.CreateShapeFromPolygon();
+                if(result.status == ActionResult.Status.Failure)
+                {
+                    m_LineMaterial.SetColor("_Highlight", Color.red);
```
thomas-tu: Please create a static field for the red color.

Editor/EditorCore/PolyShapeEditor.cs
```diff
@@ -35,8 +43,8 @@ PolyShape polygon
         Material CreateHighlightLineMaterial()
         {
             Material mat = new Material(Shader.Find("Hidden/ProBuilder/ScrollHighlight"));
-            mat.SetColor("_Highlight", new Color(0f, 200f / 255f, 170f / 200f, 1f));
-            mat.SetColor("_Base", new Color(0f, 136f / 255f, 1f, 1f));
+            mat.SetColor("_Highlight", k_LineMaterialHighlightColor);
```
karl-: Thanks for cleaning up the existing code too 👍

Runtime/MeshOperations/AppendElements.cs
karl-: It would be nice if we could pass some parameter to the triangulation function instructing it to fail on the omission of a control point, but I don't see a quick way to make that work. Is there another way we can validate that the number of points triangulated matches the input without initializating a HashSet every time? (If not that's okay, I'm just asking)

Runtime/MeshOperations/AppendElements.cs
```diff
@@ -250,9 +260,18 @@ public static ActionResult CreateShapeFromPolygon(this ProBuilderMesh mesh, ILis
...
+                if(triangleSet.Count != vertices.Length)
+                {
+                    ClearAndRefreshMesh(mesh);
```
karl-: <s>I don't think this should clear the mesh unless the operation succeeds. It should be up to the calling function to decide what to do in the case of failure.</s>

No, on second thought I think you're right. This is fine.

Runtime/MeshOperations/AppendElements.cs
vkovec: @karl- I updated it to check the distinctIndices instead of creating a hashset.

PR #165, opened 2019-06-07
Add default shaders for SRP

karl-:
Adds a default shader and material for LWRP and HDRP as samples. When imported to the project, ProBuilder will automatically use the first found pipeline compatible material as the default material.


Samples~/LWRP/.sample.json
```diff
@@ -0,0 +1,5 @@
+{
+	"displayName":"LWRP Support",
```
thomas-tu: If I remember right, the upm-ci will add entries for the samples in package.json during the build process. I don't know if that's still valid.
Otherwise, we would need to add them manually.

karl-: I tried running it through `upm-ci`, and at least locally it does not add the samples to the manifest. Added in 28178654

PR #168, opened 2019-06-11
Resetting position of UV on Box and Planar projection so that leftmos…

MartinImgSpc:
…t lowest coordinate ends up on LowerLeft anchor of UV space.

For Box projection every polygon set determined by the normal of one face of the box will have it's lower left corner aligned with the left anchor.

For planar it is the selection which will get aligned to the lower left anchor.


Editor/EditorCore/UVEditor.cs
```diff
@@ -2272,6 +2272,23 @@ Bounds2D UVSelectionBounds()
             return new Bounds2D(new Vector2((xMin + xMax) / 2f, (yMin + yMax) / 2f), new Vector2(xMax - xMin, yMax - yMin));
         }
 
+        /// <summary>
+        /// Returns the miniaml u and v values of the current selection in UV space
```
thomas-tu: Typo: `miniaml` -> `minimal`

Editor/EditorCore/UVEditor.cs
```diff
@@ -2272,6 +2272,23 @@ Bounds2D UVSelectionBounds()
...
+        /// </summary>
+        /// <returns></returns>
+        internal Vector2 UVSelectionLowerLeftUV()
+        {
+            Vector2 minimalUV = new Vector2();
+            minimalUV = Vector2.zero;
```
thomas-tu: Why not initialize directly Vector2 with Vector2.zero from declaration?

Editor/EditorCore/UVEditor.cs
```diff
@@ -3292,13 +3308,13 @@ public void Menu_FitUVs()
         }
 
         /// <summary>
-        /// Moves the selected UVs to where their bounds center is now point, where point is in UV space. Does not call ToMesh or Refresh.
+        /// Moves the selected UVs to where the anchor is now point, where both anchor_position and point are in UV space. Does not call ToMesh or Refresh.
         /// </summary>
+        /// <param name="anchor_position"></param>
         /// <param name="point"></param>
-        void CenterUVsAtPoint(Vector2 point)
+        void CenterUVsAtPoint(Vector2 anchor_position,  Vector2 point)
```
thomas-tu: `Parameter names are camelCase`:  `anchor_position` => `anchorPosition`.

Editor/EditorCore/UVEditor.cs
MartinImgSpc: Good catch! Changed!

Editor/EditorCore/UVEditor.cs
MartinImgSpc: Indeed changed

Editor/EditorCore/UVEditor.cs
MartinImgSpc: Thanks! Changed!

PR #171, opened 2019-06-13
Wb 131 enable copy uv of manual u vs

MartinImgSpc:
Allow CopyPaste of manual uv. Ensure that after a paste the destination is set to manual/auto based on the source. Added the ability to CopyPaste directly in the UV editor.


Editor/EditorCore/UVEditor.cs
```diff
@@ -718,44 +718,81 @@ void OnScreenResize()
...
 
-                ProBuilderEditor.instance.GetFirstSelectedFace(out firstObj, out source);
+                bool destinationWasManualUV = targetFace.manualUV;
+                bool sourceWasManualUV = source[0].manualUV;
+                if (sourceWasManualUV == true)
+                {
+                    //We need to convert it to auto
+                    UVEditing.SetAutoUV(firstObj, source, true);
+                    pb.Refresh();
+                    pb.Optimize();
```
karl-: There's no need to call Refresh and Optimize here. They only need to be called when we're done editing the mesh.

Editor/EditorCore/UVEditor.cs
```diff
@@ -718,44 +718,81 @@ void OnScreenResize()
...
+        bool CopyFaceUVSettings(ProBuilderMesh pb, Face targetFace)
         {
-            Event e = Event.current;
+            // get first selected UV face
+            ProBuilderMesh firstObj;
+            Face[] source = new Face[1];
 
-            // Copy UV settings
-            if (e.modifiers == (EventModifiers.Control | EventModifiers.Shift))
+            ProBuilderEditor.instance.GetFirstSelectedFace(out firstObj, out source[0]);
```
karl-: I think this should instead check `MeshSelection.activeMesh` and `MeshSelection.activeFace`.

Editor/EditorCore/UVEditor.cs
```diff
@@ -718,44 +718,81 @@ void OnScreenResize()
...
+                    //We need to convert it back to false
+                    UVEditing.SetAutoUV(firstObj, source, false); 
+                }
 
-                    selectedFace.uv = new AutoUnwrapSettings(source.uv);
-                    selectedFace.submeshIndex = source.submeshIndex;
-                    EditorUtility.ShowNotification("Copy UV Settings");
+                pb.ToMesh();
+                pb.Refresh();
+                pb.Optimize();
```
karl-: Same note as above, we only need these calls at the end of mesh edits.

Editor/EditorCore/UVEditor.cs
```diff
@@ -1150,6 +1202,21 @@ private void SetTool_Internal(Tool tool)
...
+            {
+                if (Math.PointInPolygon(pb.texturesInternal, mpos, pb.facesInternal[i].edgesInternal.AllTriangles()))
+                {
+                    faceSelected = pb.facesInternal[i];
+                    return true;
+                }
+            }
+            faceSelected = new Face();
+            return false;
+        }
```
karl-: 👍 to moving this to a method.

Editor/EditorCore/UVEditor.cs
MartinImgSpc: If I do not call those I get error
image

Editor/EditorCore/UVEditor.cs
MartinImgSpc: Changed

Editor/EditorCore/UVEditor.cs
MartinImgSpc: Same comment as above

Editor/EditorCore/UVEditor.cs
karl-: All of the ToMesh and Refresh calls can be deferred to the end of the function. I quickly tried it out here, which works: https://github.com/Unity-Technologies/com.unity.probuilder/tree/WB-131_Enable_CopyUV_OfManual_UVs_Edit

MartinImgSpc: As mentioned it did not work when the source and target manual/auto did not match. Also when source was manual.

gabrielw-us:
Problem:
1) Set Face A to Manual
2) Set Face B to Auto
3) Attempt Copy-Paste from B onto A
4) Nothing appears to happen
5) Attempt again - this time it appears to work

After much digging around, I realized that in step 3 above, the Manual face is converted to Auto, but the actual UV settings are not applied. Step 5 does that. Those two should happen in a single step (when pasted).

Copy from an Auto face to a Manual face seems to work correctly, just one step.

MartinImgSpc:
> Problem:
> 
> 1. Set Face A to Manual
> 2. Set Face B to Auto
> 3. Attempt Copy-Paste from B onto A
> 4. Nothing appears to happen
> 5. Attempt again - this time it appears to work
> 
> After much digging around, I realized that in step 3 above, the Manual face is converted to Auto, but the actual UV settings are not applied. Step 5 does that. Those two should happen in a single step (when pasted).
> 
> Copy from an Auto face to a Manual face seems to work correctly, just one step.

@gabrielw-us good catch this should be fixed now! Thank you!

MartinImgSpc:
@gabrielw-us  @karl- I have removed the conversion from the source face UV settings so that there is no loss during conversion. I think we should do another test pass to ensure there is no regression.

Runtime/MeshOperations/UV/AutoUvConversion.cs
```diff
@@ -83,14 +83,34 @@ public override string ToString()
...
+        /// </summary>
+        /// <returns></returns>
+        internal static AutoUnwrapSettings GetAutoUnwrapSettings(ProBuilderMesh mesh, Face face)
+        {
+            if (!face.manualUV)
+                return new AutoUnwrapSettings(face.uv);
+
+            var trs = GetUVTransform(mesh, face);
+            var uvSettings = AutoUnwrapSettings.defaultAutoUnwrapSettings;
+            uvSettings.offset = trs.translation;
```
karl-: doesn't the offset need to be negated when assigned to `AutoUnwrapSettings`?

Editor/EditorCore/UVEditor.cs
```diff
@@ -718,44 +718,91 @@ void OnScreenResize()
...
+            bool sourceWasManualUV = source[0].manualUV;
+                
+            if (sourceWasManualUV == true)
             {
-                // get first selected Auto UV face
-                ProBuilderMesh firstObj;
-                Face source;
+                //We need to convert it to auto
+                firstObj.ToMesh();
+                UVEditing.SetAutoUV(firstObj, source, true);
```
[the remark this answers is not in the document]
MartinImgSpc: You are looking at the outdated version

Runtime/MeshOperations/UV/AutoUvConversion.cs
MartinImgSpc: testing says otherwise

PR #173, opened 2019-06-14
macOS compatibility and minor optimizations

karl-:


MartinImgSpc:
Good catch for the OsX, however the optimization does not work in the case your source polygon is in manual UV mode. 
ErrorWithSourceConversion

PR #179, opened 2019-07-08
Fix boolean menu not respecting the Experimental Features Enabled pref

karl-:
https://fogbugz.unity3d.com/f/cases/1167692/


gabrielw-us:
Is it expected that there is a 5-10 second pause as something re-imports/recompiles/something when toggling the Experimental checkbox now? I assume yes?

karl-:
> Is it expected that there is a 5-10 second pause as something re-imports/recompiles/something when toggling the Experimental checkbox now? I assume yes?

yep, it needs to recompile scripts when changing the pref.

PR #182, opened 2019-07-16
Fix Boolean operations losing material information

karl-:
https://fogbugz.unity3d.com/f/cases/1167627/

Fixes boolean operations not retaining material information. Also moves the CSG library out of  MeshOperations and into the External directory.


External/CSG/Classes/CSG_Model.cs
```diff
@@ -0,0 +1,139 @@
...
+    /// <summary>
+    /// Representation of a mesh in CSG terms. Contains methods for translating to and from UnityEngine.Mesh.
+    /// </summary>
+    sealed class CSG_Model
+    {
+        internal List<CSG_Vertex> vertices;
+        List<Material> m_Materials;
+        List<List<int>> m_Indices;
+
+        public List<Material> materials
```
thomas-tu: This could potentially returns  null if we only create a new CSG_Model from its constructor.
Ctor instantiates `m_Indices` and `vertices` but not `m_Materials`, that seems inconsistent.

External/CSG/Classes/CSG_Model.cs
```diff
@@ -0,0 +1,139 @@
...
+            get { return m_Materials; }
+            set { m_Materials = value; }
+        }
+
+        public Mesh mesh
+        {
+            get { return (Mesh)this; }
+        }
+
+        public CSG_Model()
```
thomas-tu: I fail to see the point of this public constructor as we can't set the indices from outside (additionnal, m_Material would be null after instantiation).
The only ways I see to have an instance of CSG_Model with valid data is to use the constructor overloads.

External/CSG/Classes/CSG_Model.cs
```diff
@@ -0,0 +1,139 @@
...
+using System.Linq;
+
+namespace UnityEngine.ProBuilder.Csg
+{
+    /// <summary>
+    /// Representation of a mesh in CSG terms. Contains methods for translating to and from UnityEngine.Mesh.
+    /// </summary>
+    sealed class CSG_Model
+    {
+        internal List<CSG_Vertex> vertices;
```
thomas-tu: Is there a reason why this is a property instead of being a backing field with proper accessor? It feels inconsistent with `m_Materials` and `m_Indices` as backing fields.

External/CSG/Classes/CSG_Model.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                if (mesh.GetTopology(i) != MeshTopology.Triangles)
+                    continue;
+                var indices = new List<int>();
+                mesh.GetIndices(indices, i);
+                m_Indices.Add(indices);
+            }
+        }
+
+        internal CSG_Model(List<CSG_Polygon> list)
```
thomas-tu: `list` is generic. Maybe use `polygons`?

External/CSG/Classes/CSG_Model.cs
karl-: I just didn't bother to change it from what the library had in the first place

gabrielw-us:
Attempting to test this with simple cubes, and am getting "...caused a stack overflow" on anything other than very simple, non-scaled cubes. So the "nest operations" cannot be tested, at least I can't find a way to do it reliably.

I've proposed this before, I think we should not fix secondary Boolean issues while Boolean is still experimental. We should decide if/when there is a good sprint to dedicate to making Boolean's basic functionality reliable and predictable, and if that can be done and then confirmed good, then look into these secondary items.

gabrielw-us:
To the point of this PR - if someone can show me how you expect his to work, and make Boolean do the things required to complete the test, I'm happy to approve it, but currently I just can't get the test steps to complete :(

karl-:
I wasn't able to reproduce the issue on macOS. However, this looks like it may be a problem with the mesh normals and not necessarily the materials. In the Inspector, does the new object's MeshRenderer show one material or two?

PR #183, opened 2019-07-17
[WIP] Fix Rename Ending When a Probuilder Custom Tooltip is Shown

antoine-unity:
The code was using ShowPopup which focuses the window shown, which in turn stop any renaming. We are now using ShowPopupWithMode(ShowModeType.Popup, focus: false) which is what the newer (2019.1+) ShowTooltip does.

@gabrielw-us For tests: These change only affects 2018.3 and older.


Editor/EditorCore/TooltipEditor.cs
```diff
@@ -12,6 +13,22 @@ namespace UnityEditor.ProBuilder
     /// </summary>
     sealed class TooltipEditor : EditorWindow
     {
+#if !UNITY_2019_1_OR_NEWER
+        static TooltipEditor()
+        {
+            s_ShowModeEnum = ReflectionUtility.GetType("UnityEditor.ShowMode");
+            Assert.IsNotNull(s_ShowModeEnum, "Failed to get ShowMode enum");
```
karl-: In other places where reflection is used, we don't `Assert` or throw errors unless in Debug mode.

I think in this case we should first try to use the reflected version, but on failure, fall back to the publicly exposed `ShowPopup`.

PR #184, opened 2019-07-19
Fix vertices collapsing to a center point when "Snap on All Axes" is enabled

karl-:
Fix case where the Translate handle would snap on all axes, and then snap vertex positions on all axes. This results in incorrect vertex positions, commonly ending with all vertices smooshed into a single point.

https://unity3d.atlassian.net/browse/WB-1211


gabrielw-us:
Still having troubles, in Face mode it verts will collapse onto each other, creating a single edge (not a black hole, but similar)

gabrielw-us:
face-snap

gabrielw-us:
Still seeing snap "black hole" in this version, using:
PB git (this branch)
ProGrids 3.0.3p4
Unity 18.4LTS

snap-fail

gabrielw-us:
Sorry, not sure how I lost track of this one - still seeing the issue though:

snap-bad

PR #189, opened 2019-08-13
Wb 1398 select by material

MartinImgSpc:
Select by material was selecting by material index rather than by material. This change address this issue.

### DIFF
--- a/Editor/MenuActions/Selection/SelectMaterial.cs
+++ b/Editor/MenuActions/Selection/SelectMaterial.cs
@@ -78,20 +78,40 @@ namespace UnityEditor.ProBuilder.Actions
 
             UndoUtility.RecordSelection("Select Faces with Material");
 
-            HashSet<int> sel = new HashSet<int>(
-                    MeshSelection.topInternal
-                    .SelectMany(x => x.selectedFacesInternal.Select(y => y.submeshIndex)));
+            //Need to go from submesh index to material
+            List<Material> selectedMaterials = new List<Material>();
+            foreach(var pb in MeshSelection.topInternal)
+            {
+                HashSet<int> submeshIndex = new HashSet<int>(pb.selectedFacesInternal.Select(y => y.submeshIndex));
+                foreach(var index in submeshIndex)
+                {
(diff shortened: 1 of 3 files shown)


Editor/MenuActions/Selection/SelectMaterial.cs
```diff
@@ -78,20 +78,40 @@ public override ActionResult DoAction()
...
-                    MeshSelection.topInternal
-                    .SelectMany(x => x.selectedFacesInternal.Select(y => y.submeshIndex)));
+            //Need to go from submesh index to material
+            List<Material> selectedMaterials = new List<Material>();
+            foreach(var pb in MeshSelection.topInternal)
+            {
+                HashSet<int> submeshIndex = new HashSet<int>(pb.selectedFacesInternal.Select(y => y.submeshIndex));
+                foreach(var index in submeshIndex)
+                {
+                    if(!selectedMaterials.Contains(pb.renderer.sharedMaterials[index]))
```
karl-: Does this handle null materials?

Editor/MenuActions/Selection/SelectMaterial.cs
```diff
@@ -78,20 +78,40 @@ public override ActionResult DoAction()
...
+                    {
+                        selectedMaterials.Add(pb.renderer.sharedMaterials[index]);
```
karl-: Rather than creating two collections for submesh index and material, would it be more efficient to collect selected materials in a hashset?

Editor/MenuActions/Selection/SelectMaterial.cs
MartinImgSpc: yes

MartinImgSpc:
> Will it be possible to add tests for this behaviour please? :)

yes

Editor/MenuActions/Selection/SelectMaterial.cs
MartinImgSpc: yes, make sense!

thomas-tu:
> > Will it be possible to add tests for this behaviour please? :)
> 
> yes

Thank you!

PR #191, opened 2019-08-13
Check if there is an Inspector and if locked offer to unlock it other…

MartinImgSpc:
…wise cancel the New PolyShape creation.

image

### DIFF
--- a/Editor/MenuActions/Editors/NewPolyShape.cs
+++ b/Editor/MenuActions/Editors/NewPolyShape.cs
@@ -34,22 +34,39 @@ namespace UnityEditor.ProBuilder.Actions
 
         public override ActionResult DoAction()
         {
-            var inspWindow = EditorWindow.GetWindow<InspectorWindow>();
-            if (inspWindow != null && inspWindow.isLocked == true)
+#if UNITY_2019_1_OR_NEWER
+            //First created inspector seems to hold a specific semantic where
+            //if not unlocked no matter how many inspectors are present they will
+            //not allow the creation of new PolyShape.
+            var inspWindows = InspectorWindow.GetInspectors();
+            bool someInspectorLocked = false;
+            foreach(var insp in inspWindows)
+            {
+                if (insp.isLocked)
+                {
+                    someInspectorLocked = true;
+                    break;
+                }
+            }
+            if (someInspectorLocked == true)
             {
                 if (UnityEditor.EditorUtility.DisplayDialog(
-                                    "Inspector Locked",
-                                    "To create new Poly Shape you need access to the Inspector, which is currently locked. Do you wish to unlock the Inpsector?",
-                                    "Unlock",
-                                    "Cancel"))
+                                    L10n.Tr("Inspector Locked"),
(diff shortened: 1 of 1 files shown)


Editor/MenuActions/Editors/NewPolyShape.cs
```diff
@@ -34,6 +34,23 @@ public override bool enabled
 
         public override ActionResult DoAction()
         {
+            var inspWindow = EditorWindow.GetWindow<InspectorWindow>();
```
karl-: This will instantiate a new Inspector window if one is not already open. This should just look for an existing window.

Editor/MenuActions/Editors/NewPolyShape.cs
```diff
@@ -34,6 +34,23 @@ public override bool enabled
...
+            if (inspWindow != null && inspWindow.isLocked == true)
```
karl-: This is assuming there is only a single Inspector window open. There may be multiple, with any variation of locked and unlocked.

Editor/MenuActions/Editors/NewPolyShape.cs
```diff
@@ -34,6 +34,23 @@ public override bool enabled
...
+            {
+                if (UnityEditor.EditorUtility.DisplayDialog(
+                                    "Inspector Locked",
```
karl-: User facing text needs to be localized. Use `L10n.Tr` or `EditorGUIUtility.TrTextContent`

Editor/MenuActions/Editors/NewPolyShape.cs
MartinImgSpc: Indeed good catch.

Editor/MenuActions/Editors/NewPolyShape.cs
MartinImgSpc: I will look at all Inspector window. This will be only an issue if everything is locked.

Editor/MenuActions/Editors/NewPolyShape.cs
MartinImgSpc: Thanks! Will do!

MartinImgSpc: BTW none of the other DisplayDialog uses the localized version.

thomas-tu:
I just talked to @karl- about this PR:
- We will skip localization.
- We can use `Resources.FindObjectsOfTypeAll()` to grab the inspectors.

Editor/MenuActions/Editors/NewPolyShape.cs
```diff
@@ -32,8 +33,52 @@ public override bool enabled
             get { return true; }
         }
 
+        bool CanCreateNewPolyShape()
+        {
+            //If inspector is locked we cannot create new PolyShape.
+            //First created inspector seems to hold a specific semantic where
+            //if not unlocked no matter how many inspectors are present they will
+            //not allow the creation of new PolyShape.
+            var inspectorType = typeof(Editor).Assembly.GetType("UnityEditor.InspectorWindow");
```
karl-: This looks good for `2018.{3,4}`, though I would still use the original `InspectorWindow.GetWindows` if Unity is 2019.1 or up

PR #192, opened 2019-08-14
Adjust rotation of the PolyShape game object when using ProGrid so th…

MartinImgSpc:
…at it uses a plane normal facing the camera to adjust the rotation.


gabrielw-us:
Issue: if viewing "upward", the shape is created going into the grid/shape:

image

MartinImgSpc:
> Issue: if viewing "upward", the shape is created going into the grid/shape:
> 
> image
> 
> image

@gabrielw-us  this was the case prior to this change but it is very annoying. I have addressed it with the latest change. Thank you!

MartinImgSpc:
@thomas-tu  @karl- following the issue found by @gabrielw-us  I had to change the implementation. of CreateShapeFromPolygon. If you have a chance to review.  Thank you!

PR #194, opened 2019-08-16
Wb 1192 freeze transform

MartinImgSpc:
Freeze transform will now freeze to parent if there is a parent, otherwise world.
It also address the issue with children being moved when freezing transform of parent.

### DIFF
--- a/Editor/MenuActions/Object/FreezeTransform.cs
+++ b/Editor/MenuActions/Object/FreezeTransform.cs
@@ -27,7 +27,7 @@ namespace UnityEditor.ProBuilder.Actions
         static readonly TooltipContent s_Tooltip = new TooltipContent
             (
                 "Freeze Transform",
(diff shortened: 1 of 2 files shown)


Editor/MenuActions/Object/FreezeTransform.cs
```diff
@@ -43,27 +43,28 @@ public override ActionResult DoAction()
...
             for (int i = 0, c = selection.Count; i < c; i++)
-                positions[i] = selection[i].VerticesInWorldSpace();
-
-            for (int i = 0, c = selection.Count; i < c; i++)
-            {
+            {                
                 ProBuilderMesh pb = selection[i];
+                TransformUtility.UnparentChildren(pb.transform);
+                bool inParentSpace = (pb.transform.parent != null);
+                Vector3[] positions = pb.VerticesInParentSpace();
```
thomas-tu: Why not use `TransformUtility.TransformVertex()` and `TransformUtility.InverseTransformVertex()`?

PR #198, opened 2019-08-27
Fixed Issue Where Picking Wouldn't Work for Some Users

antoine-unity:


gabrielw-us:
Is there a fogbugz/jira with a good repro? Not sure I have that 100%.

gabrielw-us:
Yup, seems fine, just not sure if I'm testing correctly.

gabrielw-us:
@antoine-unity haaalp (see above)

karl-:
> Is there a fogbugz/jira with a good repro? Not sure I have that 100%.

We never did find a good repro for getting into this state, but if we set it manually in code we can reproduce the exact symptoms.

PR #203, opened 2019-09-10
Line billboards on Metal

karl-:
Enables line billboards on macOS running Metal like we currently have for platforms that support geometry shaders. The process is similar to how we draw vertex billboards, with some performance improvements.


Runtime/Core/MeshHandles.cs
```diff
@@ -211,13 +223,112 @@ internal static void CreatePointBillboardMesh(List<Vector3> points, Mesh mesh)
...
+            const ushort k_MaxPointCountUShort = ushort.MaxValue / 4;
+            var lineCount = mesh.edgeCount;
+
+            target.indexFormat = lineCount > k_MaxPointCountUShort
+                ? Rendering.IndexFormat.UInt32
+                : Rendering.IndexFormat.UInt16;
+
+            var vertices = mesh.positionsInternal;
+
+            Vector3[] positions = new Vector3[lineCount * 4];
```
karl-: This can be improved by keeping a static reused list and the new mesh API. I didn't do it as a part of this PR because it would require a separate path for pre-2019.3 code.

karl-:
@JoelFortin I addressed the problems with Edge mode confusing wireframe settings, as well as the problem with all edges highlighting during interaction

PR #204, opened 2019-09-18
Changed the Picking Distance

antoine-unity:
Request by @gabrielw-us 
https://unity3d.atlassian.net/browse/WB-1699


karl-:
Prior to changing this preference, I think it would be a good idea to test it with some external users. 

This makes selection behave significantly differently, and it would be good to know that this is definitive improvement for a majority of users. If nothing else, it could inform a decision to surface this as a preference if the feedback is mixed.

gabrielw-us:
Pretty straightforward actually, it's following expected behavior for Max, Blender, Maya, and general selection-related tasks. This means experienced DCC users feel at home, and for new users, there is much less "I'm not hovering over/near that thing- why is it getting selected?". That's two problems solved.

Note: Maya has a slight difference, when mouse is inside the object, some element is always highlighted, regardless of distance. That's strongly overruled by other examples though, and is not new-user-friendly. We could always add a toggle for this behavior as a separate task, but then it would also make sense to add the slider for picking distance, since the distance in each DCC is generally much smaller than the proposed 40.

PR #207, opened 2019-10-04
Fix subdivide resulting in out of bounds vertex counts

karl-:
Enable `ProBuilderMesh` to compile with `indexFormat = Uint32`, removing the previous vertex count limit of `ushort.MaxValue`.

Fixes mesh corruption when subdividing beyond 5 or 6 levels.

fogbugz: https://fogbugz.unity3d.com/f/cases/1041727/
fogbugz: https://fogbugz.unity3d.com/f/cases/1069578/

Note to reviewers - The "Connect" functions are in need of significant refactoring. Garbage is generated by the megabytes, though largely in the external Poly2Tri library. Performance can also be improved, as well as readability. For instance, "Subdivide" and "Connect Edges" / "Connect Vertices" should probably be split into separate functions to take advantage of optimizations for each specific case.


External/Poly2Tri/Triangulation/TriangulationContext.cs
```diff
@@ -36,7 +36,7 @@ abstract class TriangulationContext {
 		public TriangulationDebugContext DebugContext { get; protected set; }
 
 		public readonly List<DelaunayTriangle> Triangles = new List<DelaunayTriangle>();
-		public readonly List<TriangulationPoint> Points = new List<TriangulationPoint>(200);
+		public readonly List<TriangulationPoint> Points = new List<TriangulationPoint>();
```
antoine-unity: What's the reason for removing the preallocation of the list to a bigger size?

karl-: In the majority of cases triangulation is used for single faces, which are usually around 6 points.

PR #212, opened 2019-11-04
Disable warning for deprecation of Lightmapping API

chikuba:
https://ono.unity3d.com/unity/unity/pull-request/95782/_/lighting/lighting-settings-asset-take2 will hopefully land in 20.1 and will make all old settings API points for Lightmapping obsolete. They are still usable but will fire a warning. These changes make sure that we can still use the old API but it won't cause compile or runtime issues.
Testing status
Manual Tests: What did you do?
Nothing. It all seems to work with just disabling this warning. Have had to do this on a lot of code before.

Comments to reviewers
There is an internal API called Lightmapping.GetLightingSettingsOrDefaultsFallback() but we can't allow this to be public since we can't allow users to possibly change the global default settings that is read-only. Not sure how to do this in a nicer way for HDRPs right now.


karl-:
@chikuba is the correct API to disable automatic lightmap generation now `LightingSettings.autoGenerate`?

PR #217, opened 2019-11-13
esc always exits to no edit mode, never deletes object

karl-:
https://fogbugz.unity3d.com/f/cases/1198138/

### DIFF
--- a/Editor/EditorCore/PolyShapeEditor.cs
+++ b/Editor/EditorCore/PolyShapeEditor.cs
@@ -40,7 +40,7 @@ namespace UnityEditor.ProBuilder
             get { return target as PolyShape; }
         }
 
-        Material CreateHighlightLineMaterial()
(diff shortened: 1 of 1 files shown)


Editor/EditorCore/PolyShapeEditor.cs
```diff
@@ -40,7 +40,7 @@ PolyShape polygon
             get { return target as PolyShape; }
         }
 
-        Material CreateHighlightLineMaterial()
+        static Material CreateHighlightLineMaterial()
```
thomas-tu: How this change affects this PR?

karl-: It's not directly related, but it should be static and it seems wasteful to open a new pr for it :)

thomas-tu: Okey, just wanted to make that wasn't part of another PR :)

PR #221, opened 2019-11-27
Fix MeshFilter and MeshCollider registering as prefab property overrides

karl-:
[Fogbugz](https://fogbugz.unity3d.com/f/cases/1198568/), [Ono](https://ono.unity3d.com/unity/unity/pull-request/97300/_/st/expose-drivenpropertymanager-probuilder)

When using `ProBuilderMesh` in a Prefab, instantiating it in a scene no longer immediately marks the `MeshFilter` and `MeshCollider.mesh` properties as overridden.

In addition, the `MeshFilter` component is now marked with `HideFlags.DontSave | HideFlags.HideInInspector`, meaning it is not saved to the prefab, and the `ProBuilderMesh` component is renamed in the Inspector to `ProBuilder MeshFilter` along with the corresponding icon.

meshfilter

*Known Issues*

- Any element selection changes to an instantiated prefab still register as overrides. This will be addressed in a follow-up PR that moves the selection out of `ProBuilderMesh`.
- The `ProBuilderMesh` icon is always the Pro Skin version
- The `ProBuilderMesh` icon is toggle-able in the Gizmos window (it is off by default, however)


JoelFortin:
Right off the bat, Probuilderize action doesn't seem to work anymore.

JoelFortin:
Here's what I get trying to ProBuilderize a primitive Cube or Plane:
image

Runtime/Core/ProBuilderMesh.cs
```diff
@@ -129,8 +136,8 @@ internal new MeshRenderer renderer
         {
             get
             {
-                if (m_MeshRenderer == null)
-                    m_MeshRenderer = GetComponent<MeshRenderer>();
+                if (!gameObject.TryGetComponent<MeshRenderer>(out m_MeshRenderer))
```
antoine-unity: Is there a reason why this called isn't inside a `if (m_MeshRenderer == null)` ?

Runtime/Core/ProBuilderMeshFunction.cs
```diff
@@ -326,13 +350,14 @@ public void Refresh(RefreshMask mask = RefreshMask.All)
...
-
-            if (meshCollider != null)
+            if(gameObject.TryGetComponent<MeshCollider>(out collider))
             {
-                gameObject.GetComponent<MeshCollider>().sharedMesh = null;
-                gameObject.GetComponent<MeshCollider>().sharedMesh = mesh;
+                SerializationUtility.UnregisterDrivenProperty(this, collider, "m_Mesh");
+                SerializationUtility.RegisterDrivenProperty(this, collider, "m_Mesh");
+                collider.sharedMesh = null;
+                collider.sharedMesh = mesh;
```
antoine-unity: Is this double assignation required?

Runtime/Core/ProBuilderMesh.cs
karl-: `m_MeshRenderer` can be deleted, this change was intended to make the `renderer` property no longer cached.

Runtime/Core/ProBuilderMeshFunction.cs
karl-: It used to be a bug that this was required, it might no longer be necessary. I will check.

PR #224, opened 2019-12-16
Fixed Issue Where Objects Could Be Selected Through Popups

antoine-unity:
It was possible to select an object when double clicking in the object selector with a object in the scene view behind. 

[[case #1161998]](https://fogbugz.unity3d.com/f/cases/1161998/)


Editor/EditorCore/ProBuilderEditor.cs
```diff
@@ -705,8 +707,10 @@ void OnSceneGUI(SceneView sceneView)
                     m_WasDoubleClick = false;
             }
 
-            if (m_CurrentEvent.type == EventType.MouseUp)
+            if (m_CurrentEvent.type == EventType.MouseUp && GUIUtility.hotControl == m_DefaultControl)
             {
+                GUIUtility.hotControl = 0;
+
```
karl-: I think we will also need to relinquish the hot control on `Ignore` or `Escape` events, and anywhere else that `m_IsDragging` is unset.

antoine-unity: Is Escape currently cancelling a drag in probuilder?

karl-: If it doesn't, it should 😝

PR #226, opened 2020-01-06
Fix errors when assigning null mesh to preview shape

karl-:
https://fogbugz.unity3d.com/f/cases/1203585/

Fixes an issue where assigning `null` to the `New Shape` editor preview object would throw errors.


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [4.3.0-preview.2] - 2020-01-06
+
+## Bug Fixes
+
+- Fixed the `Custom Shape` field in the `Shape Editor` throwing errors on invalid input.
```
JoelFortin: This should only state that we remove the Custom Shape field now.

PR #233, opened 2020-01-07
Update documentation link in Smoothing panel

thomas-tu:
**Goal of this PR:**
Fix this case: https://fogbugz.unity3d.com/f/cases/1183101

### DIFF
--- a/Editor/EditorCore/SmoothGroupEditor.cs
+++ b/Editor/EditorCore/SmoothGroupEditor.cs
@@ -507,7 +507,7 @@ namespace UnityEditor.ProBuilder
                 global::UnityEditor.ProBuilder.UI.EditorGUILayout.EndRow();
 
                 if (GUILayout.Button("Open Documentation"))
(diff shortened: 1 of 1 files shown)


Editor/EditorCore/SmoothGroupEditor.cs
```diff
@@ -507,7 +507,7 @@ void OnGUI()
                 global::UnityEditor.ProBuilder.UI.EditorGUILayout.EndRow();
 
                 if (GUILayout.Button("Open Documentation"))
-                    Application.OpenURL("http://procore3d.github.io/probuilder2/toolbar/tool-panels/#smoothing-groups");
+                    Application.OpenURL("https://docs.unity3d.com/Packages/com.unity.probuilder@4.2/manual/workflow-edit-smoothing.html");
```
karl-: Is there a `@latest` or similar that we can use instead of hard-coding 4.2?

thomas-tu: Yes, it is done in 7d05b91

PR #234, opened 2020-01-08
added case for URP/HDRP shaders

neilunity:
fix for https://fogbugz.unity3d.com/f/cases/1205318/
Legacy uses _Color keyword
HDRP uses _BaseColor and _Color
URP uses _BaseColor

rather dependant on how shaders are written

### DIFF
--- a/Editor/EditorCore/ObjExporter.cs
+++ b/Editor/EditorCore/ObjExporter.cs
@@ -351,10 +351,17 @@ namespace UnityEditor.ProBuilder
                     }
                 }
 
-                if (mat.HasProperty("_Color"))
+                Color color = mat.color;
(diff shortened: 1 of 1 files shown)


Editor/EditorCore/ObjExporter.cs
```diff
@@ -351,10 +351,17 @@ static string WriteMtlContents(IEnumerable<Model> models, ObjOptions options, ou
                     }
                 }
 
-                if (mat.HasProperty("_Color"))
+                Color color = mat.color;
+                if (mat.HasProperty("_BaseColorMap") || mat.HasProperty("_BaseMap")) // HDRP || URP
+                {
+                    color = mat.GetColor("_BaseColor");
```
karl-: Does `"_BaseMap"` also need to be looked at here?

neilunity: this how it is done in the model exporter code

PR #237, opened 2020-01-08
Create render custom pass in our Picker to better support HDRP

thomas-tu:
**Goal of this PR:**
With HDRP, the way we do our picking pass must change as it brings few issues with our lookup texture:
- it is vertically flipped
- gizmos and colliders are visible and can overwrite our objects

Creating our own render pass based on ScriptableRenderContext is a better way to handle this and gives us full control of what we render.

Fix this case: https://fogbugz.unity3d.com/f/cases/1204731


Runtime/Unity.ProBuilder.asmdef
```diff
@@ -2,15 +2,22 @@
     "name": "Unity.ProBuilder",
     "references": [
         "Unity.ProBuilder.Poly2Tri",
-        "Unity.ProBuilder.KdTree"
+        "Unity.ProBuilder.KdTree",
+        "Unity.RenderPipelines.HighDefinition.Runtime"
```
karl-: Does this also need to reference Universal Render Pipeline, or is that implicit?

thomas-tu: Not at the moment. This PR is solely for HDRP.
The old codepath still work with URP so I kept the way it is. At some point, we'll have to give it a deeper look.

thomas-tu: After a quick look, URP doesn't really work. It seems to work but in many cases it won't work as we draw way more on the lookup texture than just our picking data. We filed a ticket for that.

Tests/Editor/Picking/RectSelectionPicker.cs
```diff
@@ -0,0 +1,138 @@
...
+/// Issues will likely to appear if changes have been made in graphics pipeline.
+/// This set of tests should run against standard pipeline, URP and HDRP.
+/// </summary>
+class RectSelectionPicker
+{
+    ProBuilderMesh[] selectables;
+    GameObject[] sceneObjects;
+    Camera camera;
+
+    void Setup()
```
antoine-unity: Why is this not using [SetUp] test attribute?

Tests/Editor/Picking/RectSelectionPicker.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+        };
+
+#if UNITY_2019_1_OR_NEWER
+        foreach (SceneView sv in SceneView.sceneViews)
+            sv.drawGizmos = true;
+#endif
+    }
+
+    void Cleanup()
```
antoine-unity: Why is this not using [TearDown] test attribute?

Runtime/Core/SelectionPickerRenderer.cs
```diff
@@ -1,6 +1,5 @@
-//#define PB_RENDER_PICKER_TEXTURE
+#define PB_RENDER_PICKER_TEXTURE
```
karl-: This should not be enabled

Runtime/Core/SelectionPickerRendererHDRP.cs
```diff
@@ -0,0 +1,101 @@
...
+                img.Apply();
+
+                RenderTexture.active = prev;
+                RenderTexture.ReleaseTemporary(rt);
+
+                UObject.DestroyImmediate(go);
+
+                return img;
+#else
+                return null;
```
karl-: So if a user has a version of HDRP lower than 7.1, what happens? Does this know to fall back to legacy rendering somewhere?

Runtime/Core/SelectionPickerRendererStandard.cs
```diff
@@ -0,0 +1,109 @@
...
+                int height = -1)
+            {
+                bool autoSize = width < 0 || height < 0;
+
+                int _width = autoSize ? (int)camera.pixelRect.width : width;
+                int _height = autoSize ? (int)camera.pixelRect.height : height;
+
+                GameObject go = new GameObject();
+                Camera renderCam = go.AddComponent<Camera>();
+                renderCam.CopyFrom(camera);
```
karl-: There's a lot of camera setup that looks like it could be abstracted between all picking renderers, would it make sense to do so?

Tests/Editor/Picking/RectSelectionPicker.cs
thomas-tu: Because I followed the way it's being done in other test. I will change it.

Tests/Editor/Picking/RectSelectionPicker.cs
thomas-tu: See comment above, will rectify

thomas-tu: Fix in 085a884

Tests/Editor/Picking/RectSelectionPicker.cs
thomas-tu: Fix in 085a884

Runtime/Core/SelectionPickerRenderer.cs
thomas-tu: Fix in 085a884

Runtime/Core/SelectionPickerRendererHDRP.cs
thomas-tu: `SelectionPickerRenderer` has its own property to fetch the appropriate picker according at what pipeline is active in the project: `SelectionPickerRenderer.pickerRenderer`. If HDRP doesn't exist, it will always fallback to the standard solution.

Runtime/Core/SelectionPickerRendererStandard.cs
thomas-tu: I'm not sure, setup could evolve differently between pipelines that I chose to put a hard line rather than using a common base. I prefer to keep things separate in case we have to deal we specifics field/setup/data/etc..

PR #239, opened 2020-01-09
fixed duplicate names when creating shapes

neilunity:


neilunity:
https://fogbugz.unity3d.com/f/cases/1209864/

JoelFortin:
Also, creating a cube with the shape tool ... the very first cube created will get named "Cube (1)" since the preview object had "Cube" and the method caught it.

neilunity:
fixed

PR #244, opened 2020-01-10
Ignore changes from a Reset when users trigger it

thomas-tu:
**Goal of this PR:**
This PR prevents user to manually reset a ProBuilderMesh component by ignoring it. Otherwise, Reset would put the instance in a weird state without offering a way to populate the mesh data. That is something that users should not trigger.

Fix this case: https://fogbugz.unity3d.com/f/cases/1196134/

### DIFF
--- a/Runtime/Core/ProBuilderMeshFunction.cs
+++ b/Runtime/Core/ProBuilderMeshFunction.cs
@@ -51,6 +51,15 @@ namespace UnityEngine.ProBuilder
                 Rebuild();
         }
 
+#if UNITY_EDITOR
+        void Reset()
+        {
+            // Prevent reset from Unity Editor as it would put this component in a weird state.
+            if (meshSyncState != MeshSyncState.Null)
+                UnityEditor.Undo.PerformUndo();
+        }
+#endif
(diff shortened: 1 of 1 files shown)


Runtime/Core/ProBuilderMeshFunction.cs
```diff
@@ -51,6 +51,15 @@ void Awake()
                 Rebuild();
         }
 
+#if UNITY_EDITOR
+        void Reset()
+        {
+            // Prevent reset from Unity Editor as it would put this component in a weird state.
+            if (meshSyncState != MeshSyncState.Null)
+                UnityEditor.Undo.PerformUndo();
```
karl-: `MeshSyncState.Null` just means that the `MeshFilter.sharedMesh` is not set, not that the `ProBuilderMesh` data is invalid. In this case I think just preventing reset entirely is the way.

thomas-tu: Reset gets called between Awake and Start when you generate an object. Is there something I can use to check if the data is valid?

PR #250, opened 2020-01-27
fixed rotation clamping

neilunity:
https://fogbugz.unity3d.com/f/cases/1210096/

### DIFF
--- a/Editor/EditorCore/AutoUVEditor.cs
+++ b/Editor/EditorCore/AutoUVEditor.cs
@@ -480,6 +480,10 @@ namespace UnityEditor.ProBuilder
                 {
                     var uv = q.uv;
                     uv.rotation = rot;
+                    if (uv.rotation > 360f)
+                        uv.rotation = uv.rotation % 360f;
+                    else if (uv.rotation < 0f)
+                        uv.rotation = 360f + (uv.rotation % 360f);
                     q.uv = uv;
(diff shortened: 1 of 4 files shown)


Editor/EditorCore/AutoUVEditor.cs
```diff
@@ -480,6 +480,10 @@ private static void SetRotation(float rot, ProBuilderMesh[] sel)
                 {
                     var uv = q.uv;
                     uv.rotation = rot;
+                    if (uv.rotation > 360f)
+                        uv.rotation = uv.rotation % 360f;
+                    else if (uv.rotation < 0f)
+                        uv.rotation = 360f + (uv.rotation % 360f);
```
karl-: Should this be `360f - (uv.rotation % 360f)`?

Editor/EditorCore/UVEditor.cs
```diff
@@ -1535,6 +1535,10 @@ void RotateTool()
                         {
                             var uv = face.uv;
                             uv.rotation += uvRotation - t_uvRotation;
+                            if (uv.rotation > 360f)
+                                uv.rotation = uv.rotation % 360f;
+                            else if (uv.rotation < 0f)
+                                uv.rotation = 360f + (uv.rotation % 360f);
```
karl-: Same note, is the intent to clamp between 0 and 360, or just prevent negative values?

PR #256, opened 2020-02-07
Fix probuilderize layout error

karl-:
Fixes an issue where the toolbar layoutting could throw an error if ProBuilderize action utilizes a progress bar.

https://fogbugz.unity3d.com/f/cases/1195261/


JoelFortin:
this fixes the stack overflow and the potential crash we were getting ... but now, just wondering if we should simply revert the operation when it happens. For now, there is a warning but we still give what the probuilderize operation was able to do, which is most likely unusable.

Also, would there be a way to give more info on what happened, like what I should do to get my model correctly probuilderize or something ?

karl-:
In the event that ProBuilderize fails it _should_ do nothing to the input mesh. In my tests it doesn't do anything on failure, but I also wasn't able to reproduce the stack overflow on my computer. If it comes up again I think a separate bug should be opened for preserving the state of input when an action fails.

karl-:
@JoelFortin I didn't see that you hadn't approved this yet and accidentally merged. I reverted the merge but this PR still shows as closed, let me know if it is good to merge after my comment.

PR #262, opened 2020-02-20
Fix Poly Shape Undo Crash

karl-:
https://fogbugz.unity3d.com/f/cases/1221121/


JoelFortin:
Since we're here ... would there be a way to delete the "Add Polygon Shape Point" undo stack entries once the object is created. The thing is that those actions are still in the stack but don't have any effect, so I have to undo let's say 5 times (pass through all the points I created) before actually undoing the create action.

PR #264, opened 2020-02-28
Fix PolyShape Undo Errors

karl-:
https://fogbugz.unity3d.com/f/cases/1223330/

### DIFF
--- a/Editor/EditorCore/EditorUtility.cs
+++ b/Editor/EditorCore/EditorUtility.cs
@@ -309,11 +309,14 @@ namespace UnityEditor.ProBuilder
             switch (s_ColliderType.value)
             {
                 case ColliderType.BoxCollider:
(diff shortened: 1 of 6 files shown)


Editor/EditorCore/EditorUtility.cs
```diff
@@ -309,11 +309,14 @@ internal static void InitObject(ProBuilderMesh pb)
             switch (s_ColliderType.value)
             {
                 case ColliderType.BoxCollider:
-                    pb.gameObject.AddComponent<BoxCollider>();
+                    if(!pb.gameObject.TryGetComponent<BoxCollider>(out _))
```
neilunity: why not use getComponent ?

karl-: TryGetComponent doesn't allocate if there is no component matching the requested type. It's a minor perf. optimization.

PR #275, opened 2020-03-19
fix mesh colliders not having sharedmesh property populated in builds

karl-:
https://fogbugz.unity3d.com/f/cases/1225223/


karl-:
@JoelFortin No I just committed a fix for this at 52dc37a

PR #277, opened 2020-03-23
Fix "New Shape" menu action always building a cube

karl-:
The `New Shape` menu item now creates the last selected shape with it's last set parameters instead of always building a cube primitive with the default values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Bug Fixes
+
+- Fixed `New Shape` menu item always creating a Cube instead of the last selected shape.
+
 ## [4.3.0-preview.6] - 2020-03-23
 
 ### Bug Fixes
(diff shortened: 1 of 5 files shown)


JoelFortin:
Looks good but some questions arise for me :
1. Now that the New Button takes whatever settings we left in the shape tool, should we also map the shortcut CTRL+K to do the same ? right now it still create a cube, so we have discrepancy between the two ways now.
2. Now that I had to play with different shapes in short period of time, I would kill just to have a reset button for shape settings only in the shape window (there is a new reset option in the preferences but it's to high level and can reset options that I don't necessarily want to reset if I only need to wipe the shape data).
3. Also noted if I have the shape tool opened and I perform a reset in the preferences, I see the update right away in the shape window but the preview object in the scene is still the one I had before the reset. 

#1 should be addressed in this PR if we decide to match to behavior.
#2 is pretty much a feature request and should be treated separately

What do you think ?

karl-:
> Now that the New Button takes whatever settings we left in the shape tool, should we also map the shortcut CTRL+K to do the same ? right now it still create a cube, so we have discrepancy between the two ways now.

Cmd+K is mapped to the menu item `Create ProBuilder Cube`, so I think it makes sense that it remains as is. We don't have a shortcut for 'New Shape' by default, but one could be added if you think it's necessary here.

> Now that I had to play with different shapes in short period of time, I would kill just to have a reset button for shape settings only in the shape window (there is a new reset option in the preferences but it's to high level and can reset options that I don't necessarily want to reset if I only need to wipe the shape data).

+1, added in bed79c349f263a9d16a3805e59e437359173e93f

> Also noted if I have the shape tool opened and I perform a reset in the preferences, I see the update right away in the shape window but the preview object in the scene is still the one I had before the reset.

I wasn't able to find a good way to do this without modifying Settings Manager to provide a callback when things are reset. I will do that as future work.

PR #281, opened 2020-04-14
Bugfix/exception when stripping

neilunity:

### Purpose of this PR

Smoothing group editor was not refreshed properly after stripping pb components.

### Links

https://fogbugz.unity3d.com/f/cases/1237636/

### Comments to Reviewers


Editor/EditorCore/SmoothGroupEditor.cs
```diff
@@ -721,5 +721,13 @@ static Color32 GetDistinctColor(int index)
             new Color32(255, 255, 255, 255),    // White
             new Color32(0, 0, 0, 255),          // Black
         };
+
+        public static void Refresh()
```
karl-, line 725: I'd use the existing `OnSelectionChanged` here, replacing the `Selection.selectionChanged += OnSelectionChanged;` earlier with the ProBuilder version - `MeshSelection.objectSelectionChanged`. I'm not sure that it is invoked when the strip action runs, although if it isn't it should be.

Editor/EditorCore/StripProBuilderScripts.cs
```diff
@@ -61,6 +61,7 @@ public static void Strip(ProBuilderMesh[] all)
             UnityEditor.EditorUtility.DisplayDialog("Strip ProBuilder Scripts", "Successfully stripped out all ProBuilder components.", "Okay");
 
             ProBuilderEditor.Refresh();
+            SmoothGroupEditor.Refresh();
```
karl-, line 64: As noted above, we can avoid manually refreshing from this script by ensuring that the `MeshSelection.objectSelectionChanged` callback is invoked when scripts are removed. That way any other systems that are looking at `ProBuilderMesh` components will be notified as well.

PR #282, opened 2020-04-15
return failure if connect fails

neilunity:

### Purpose of this PR

Now return gracefully when trying to connect edges in an invalid face. Would throw an exception when subdividing an invalid shape.

### Links

https://fogbugz.unity3d.com/f/cases/1232389/

### Comments to Reviewers


Runtime/MeshOperations/ConnectElements.cs
```diff
@@ -258,23 +258,35 @@ public static int[] Connect(this ProBuilderMesh mesh, IList<int> indexes)
                         usedTextureGroups.Add(newTextureGroupIndex);
                     }
 
-                    foreach (ConnectFaceRebuildData c in res)
+
+                    if (res == null)
+                    {
+                        connections = null;
+                        addedFaces = null;
+                        return new ActionResult(ActionResult.Status.Failure, "Unable to connect faces");
```
JoelFortin, line 266: How is this warning supposed to be shown in the editor exactly ? Now when I subdivide an invalid shape, I get nothing at all ... and if I turn ON the Notification from the PB preferences windows ... I only get "Subdivide 0 Objects" notification.
Did I miss something ?

neilunity: updated message

PR #285, opened 2020-05-06
Fix polyshape actions

RoxanneCor:

### Purpose of this PR

Fixed NullException with Strip ProBuilder action on PolyShape.
Now also removing the PolyShape script when stripping ProBuilder scripts
Also added tests for this feature

### Links

https://fogbugz.unity3d.com/f/cases/1242884/

### Comments to Reviewers


Editor/EditorCore/PolyShapeEditor.cs
```diff
@@ -219,6 +219,10 @@ public void RebuildPolyShapeMesh(bool vertexCountChanged = false)
             if (polygon == null)
                 return;
 
+            // If strip ProBuilder scripts is called when selected this can occur
+            if (polygon.mesh == null)
+                return;
```
karl-, line 224: This works, but I think the complete fix here would include also removing the `PolyShape` component when stripping scripts.

Tests/Editor/Actions/StripProBuilderScripts.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        UVEditor.MenuOpenUVEditor();
+
+        m_cube = ShapeGenerator.CreateShape(ShapeType.Cube);
+        m_cube_obj = m_cube.gameObject;
+        UnityEditor.ProBuilder.EditorUtility.InitObject(m_cube);
+
+        Assume.That(m_cube_obj.GetComponent<ProBuilderMesh>() != null);
+        Assume.That(m_cube_obj.GetComponent<PolyShape>() != null);
```
karl-: Good use of `Assume`

Tests/Editor/Actions/StripProBuilderScripts.cs
```diff
@@ -0,0 +1,48 @@
...
+    public void Strip_ProBuilder_Scripts()
+    {
+        // make sure the ProBuilder window is open
+        if (ProBuilderEditor.instance == null)
+        {
+            ProBuilderEditor.MenuOpenWindow();
+            m_OpenedWindow = true;
+        }
+
+        UVEditor.MenuOpenUVEditor();
```
karl-: What's the reason for opening the Editor and UV windows?

Tests/Editor/Actions/StripProBuilderScripts.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        m_cube = ShapeGenerator.CreateShape(ShapeType.Cube);
+        m_cube_obj = m_cube.gameObject;
+        UnityEditor.ProBuilder.EditorUtility.InitObject(m_cube);
+
+        Assume.That(m_cube_obj.GetComponent<ProBuilderMesh>() != null);
+        Assume.That(m_cube_obj.GetComponent<PolyShape>() != null);
+
+        EditorApplication.ExecuteMenuItem("Tools/" + PreferenceKeys.pluginTitle + "/Actions/Strip All ProBuilder Scripts in Scene");
```
karl-: Usually when writing tests it's best to keep the scope focused. In this case, we're testing the whole pipeline from `Menu Item` to the actual `Strip Script` function. So if this test fails it could be that the menu item function failed to find any `ProBuilderMesh` components in the scene, or that the stripping function itself failed to remove the proper components.

I'd suggest calling the `StripProBuilderScripts.Strip(ProBuilderMesh[] all)` function directly to avoid any ambiguity in the results, then maybe write an additional test that verifies the menu item does collect all meshes in the scene. I think for this PR however just testing the `Strip` function is fine.

Tests/Editor/Actions/StripProBuilderScripts.cs
```diff
@@ -0,0 +1,48 @@
...
+        Assert.That(m_cube_obj.GetComponent<PolyShape>() == null);
+
+
+        // close editor window if we had to open it
+        if (m_OpenedWindow && ProBuilderEditor.instance != null)
+        {
+            ProBuilderEditor.instance.Close();
+        }
+
+        UObject.DestroyImmediate(m_cube_obj);
```
karl-: Good cleaning up after tests 👍

Tests/Editor/Actions/StripProBuilderScripts.cs
```diff
@@ -0,0 +1,48 @@
...
+using UnityEditor.ProBuilder;
+using UnityEngine;
+using UnityEngine.ProBuilder;
+using UnityEditor;
+
+public class StripProBuilderScripts
+{
+    bool m_OpenedWindow = false;
+    ProBuilderMesh m_cube;
+    GameObject m_cube_obj;
```
karl-: In general I'd suggest declaring variables in the smallest possible scope. Since these are only used in a single function, there's no need to declare them at a class level. If for example you were to reuse these in multiple functions in this class, possibly with a `[SetUp]` and `[TearDown]` then it would make sense to keep them at this scope.

Tests/Editor/Actions/StripProBuilderScripts.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    [Test]
+    public void Strip_ProBuilder_Scripts()
```
karl-, line 13: Test names should at a glance indicate what they are testing. In this case I understand that it is validating some behaviour for the script stripping function, but it's unclear exactly what. Consider `StripProBuilderScripts_RemovesMeshAndPolyShapeComponents` or `StripScripts_RemovesAllProBuilderComponents`?

Probuilder isn't the best role model in this regard at the moment :)

Tests/Editor/Actions/StripProBuilderScripts.cs
RoxanneCor: I now call DoStrip() on the object instead to avoid calls from the Editor (Strip had some UnityEditor code)

Editor/EditorCore/StripProBuilderScripts.cs
```diff
@@ -80,6 +80,9 @@ public static void DoStrip(ProBuilderMesh pb)
 
                 EditorUtility.SynchronizeWithMeshFilter(pb);
 
+                if (go.GetComponent<PolyShape>())
+                    DestroyImmediate(go.GetComponent<PolyShape>());
```
antoine-unity, line 84: You should use a variable to skip searching for the component twice. Something like this instead:
```
var polyShape = go.GetComponent<PolyShape>();
if (polyShape != null)
   DestroyImmediate(polyShape);
```

In this case, we also have a quick version of this that doesn't allocate if the component doesn't exist:
```
if (go.TryGetComponent(out PolyShape polyShape)
    DestroyImmediate(polyShape);
```
You can find the doc here: https://docs.unity3d.com/ScriptReference/GameObject.TryGetComponent.html

Tests/Editor/Actions/StripProBuilderScripts.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        var go = new GameObject();
+        var poly = go.AddComponent<PolyShape>();
+        go.AddComponent<ProBuilderMesh>();
+        poly.CreateShapeFromPolygon();
+
+        Assume.That(go.GetComponent<ProBuilderMesh>() != null);
+        Assume.That(go.GetComponent<PolyShape>() != null);
+
+        UnityEditor.ProBuilder.Actions.StripProBuilderScripts.DoStrip(go.GetComponent<ProBuilderMesh>());
```
antoine-unity, line 20: Add a `using UnityEditor.ProBuilder.Actions` instead of writing it out for more readable code.

RoxanneCor: I used this because I names both classes the same way (oups), so renamed the Test class

Editor/EditorCore/StripProBuilderScripts.cs
RoxanneCor: Since all the function was made this way I changed it in other parts too to make it more uniform and efficient :)

PR #291, opened 2020-05-19
Miscellaneous Small Fixes

karl-:
### Purpose of this PR

- [case: https://fogbugz.unity3d.com/f/cases/1248708/] Fixed physx error when welding all vertices to a singularity.
- Fixed tooltips always clamping to the left of the screen on secondary monitors.
- Fixed `EditorUtility.SetIconEnabled` not respecting the enabled parameter.
- Reduced the amount of repaints in ProBuilder window.


Editor/EditorCore/EditorToolbar.cs
```diff
@@ -424,13 +424,28 @@ public void OnGUI()
...
+
+            m_WantsRepaint = false;
+        }
+
+        void ScheduleRepaint()
+        {
+            m_WantsRepaint = true;
+        }
+
+        void ImmediateRepaint()
```
antoine-unity, line 446: Not sure about this name, at first I assumed it was RepaintImmediately and realised it was something else.

PR #293, opened 2020-05-20
Feature pathfinding faces

RoxanneCor:
### Purpose of this PR

Added a path-finding feature to select faces in the shortest path between 2 given faces

### Comments to Reviewers

Another pull request will also add new tests on the caching and a feature to preview the path before clicking.
How to test: 
1. Select a face
2. Hold ctrl+shift or cmd+shift
3. Select a second face

A path between those 2 faces should be selected


Editor/MenuActions/Selection/SelectPathFaces.cs
```diff
@@ -0,0 +1,157 @@
...
+                visited.Add(current);
+                toVisit.Remove(current);
+            } 
+
+            return predecessors;
+        }
+
+        private static float GetWeight(int face1, int face2, ProBuilderMesh mesh)
+        {
+            float baseCost = 10f;
```
RoxanneCor, line 118: The number in this part are arbitrary and will need to be changed after some real case testing

Editor/MenuActions/Selection/SelectPathFaces.cs
```diff
@@ -0,0 +1,157 @@
...
+                    {
+                        toVisit.Add(idx);
+                    }
+
+                    otherWing = otherWing.next;
+
+                } while (otherWing != currentWing);
+
+                // Look for the next face to visit, choosing the one with less weight
+                double min = double.MaxValue;
```
karl-, line 100: Is there a reason for double precision here, or can this be float?

Editor/MenuActions/Selection/SelectPathFaces.cs
```diff
@@ -0,0 +1,157 @@
...
+            } 
+
+            return predecessors;
+        }
+
+        private static float GetWeight(int face1, int face2, ProBuilderMesh mesh)
+        {
+            float baseCost = 10f;
+            float normalMult = 2f;
+            float distMult = 1f;
```
karl-, line 120: It looks like these could be `const`. It would also be good to comment on their purpose and what "good" ranges for each are.

Tests/Editor/Performance/PathToolTests.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    [TestCase("pb_path_test_large", 1, 2, Description = "Find a path between 2 faces on a large mesh, with close faces")]
+    [TestCase("pb_path_test_large", 1, 18570, Description = "Find a path between 2 faces on a large mesh, with far faces")]
+    [Test, Performance]
+    public void PathToolTest(string path, int start, int end)
+    {
+        var go = Resources.Load<GameObject>(path);
+        Assume.That(go != null);
+        var mesh = go.GetComponent<ProBuilderMesh>();
+        Assume.That(mesh != null);
```
karl-, line 27: Good use of `Assume` here

Editor/MenuActions/Selection/SelectPathFaces.cs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                weights[i] = float.MaxValue;
+            }
+
+            int current = start;
+            weights[current] = 0;
+            visited.Add(current);
+
+            // Construct the paths between the start face and every other faces
+            while (visited.Count < wingCount)
```
karl-, line 70: Does this account for the cases where geometry is non-manifold? Ex, three faces referencing sharing a common edge, or open edges?

Tests/Editor/Unity.ProBuilder.Editor.Tests.asmdef
```diff
@@ -5,18 +5,30 @@
...
         "Unity.ProBuilder.Editor",
         "Unity.ProBuilder.Stl",
-        "Unity.Settings.Editor"
-    ],
-    "optionalUnityReferences": [
-        "TestAssemblies"
+        "Unity.Settings.Editor",
+        "Unity.PerformanceTesting",
+        "UnityEngine.TestRunner",
+        "UnityEditor.TestRunner"
```
karl-, line 11: I'm not sure that these changes will compile pre-2019.1.

Editor/EditorCore/EditorSceneViewPicker.cs
```diff
@@ -6,6 +6,7 @@
 using PHandleUtility = UnityEngine.ProBuilder.HandleUtility;
 using UHandleUtility = UnityEditor.HandleUtility;
 using RaycastHit = UnityEngine.ProBuilder.RaycastHit;
+using UnityEditor.ProBuilder.Actions;
```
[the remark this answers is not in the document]
RoxanneCor: Maybe I could put the SelectPath feature in another namespace? Since it's not exactly an actio. Do you have any ideas?

Editor/MenuActions/Selection/SelectPathFaces.cs
```diff
@@ -0,0 +1,157 @@
...
+
+namespace UnityEditor.ProBuilder.Actions
+{
+    public static class SelectPathFaces
+    {
+        private static int[] lastPredecessors;
+        private static int lastStart;
+        private static ProBuilderMesh lastMesh;
+        private static List<WingedEdge> lastWings;
+        private static Dictionary<Face, int> lastFacesIndex = new Dictionary<Face, int>();
```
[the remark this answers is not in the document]
neilunity: I am not a big fan of the 'last' prefix, tends to imply last before something, and the something is not clear from the code. Rename to 'cached' or something equivalent. Clarify in code when the cache is updated.

Editor/EditorCore/EditorSceneViewPicker.cs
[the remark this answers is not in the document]
karl-: Usually what I'd do is separate the menu and UI code from the selection code. So move the actual path finding action into it's own class in the runtime API, then call into it and set the selection in your editor side action class.

Editor/MenuActions/Selection/SelectPathFaces.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        private static int[] Dijkstra(int start, int end, ProBuilderMesh mesh)
+        {
+            HashSet<int> visited = new HashSet<int>();
+            HashSet<int> toVisit = new HashSet<int>();
+            lastWings = lastMesh == mesh ? lastWings : WingedEdge.GetWingedEdges(mesh, true);
+            int wingCount = lastWings.Count;
+
+            if (mesh != lastMesh)
```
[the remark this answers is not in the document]
RoxanneCor: Do you think checking for the number of faces in the mesh could be enough? Or should I really check for a complete geometry change?

Editor/MenuActions/Selection/SelectPathFaces.cs
RoxanneCor: I'm really not sure about what ranges to use yet, but as soon as we get some testing I'll be ure to add it

Tests/Editor/Unity.ProBuilder.Editor.Tests.asmdef
RoxanneCor: Do I need to push that file? My code takes account of whether or not we have those packages anyway

Editor/MenuActions/Selection/SelectPathFaces.cs
RoxanneCor: It didn't but now it does!

Editor/MenuActions/Selection/SelectPathFaces.cs
```diff
@@ -0,0 +1,157 @@
...
+        private static Dictionary<Face, int> lastFacesIndex = new Dictionary<Face, int>();
+
+        /// <summary>
+        /// Calculates the indexes of all faces in the shortest path between start and end
+        /// </summary>
+        /// <param name="start">The index of the starting face</param>
+        /// <param name="end">The index of the ending face</param>
+        /// <param name="mesh">The mesh of the object</param>
+        /// <returns>The indexes of all faces </returns>
+        public static List<int> GetPath(int start, int end, ProBuilderMesh mesh)
```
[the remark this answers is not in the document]
RoxanneCor: I added some exceptions but I rarely worked with those so I'm not sure I'm using them well?

Tests/Editor/Unity.ProBuilder.Editor.Tests.asmdef
karl-: Yes, you will need to update the asmdef with the optional dependencies.  Looking at yamato it seems like the point of failure is the NUnit dll not being found - https://yamato-artifactviewer.prd.cds.internal.unity3d.com/0e7f19c6-06b4-4b68-9dad-38fc9a7068ee%2Flogs%2Fupm-ci~%2Ftest-results%2Fisolation-2265397-com.unity.probuilder/TestReportV1.html

You might try leaving your changes but including the
```
"optionalUnityReferences": [ "TestAssemblies" ]
```
key from the original.

gabrielw-us:
Noticed that "SHIFT" click will not set active face in this branch. Is that fixed elsewhere?

gabrielw-us:
Also:
1. Click one face
2. SHIFT-click another face
3. CTRL click the first face

Expected: face is deselected
Result: face remains selected

gabrielw-us:
Oh shoot I'm sorry, in the wrong branch :D

gabrielw-us:
Spamalam - ok, in the correct branch, still seeing those issues :(

PR #294, opened 2020-05-27
Fix shift select

RoxanneCor:
### Purpose of this PR

Fixed Shift selection being inconsistent

### Links

https://fogbugz.unity3d.com/f/cases/1249071/

### Comments to Reviewers

There is the same issue with edges and vertexes, so if the code for the PR is okay I will fix the other 2 in this same PR.

EDIT: Added code to fix vertexes and edges as well. Pretty much the same as with faces.

### DIFF
--- a/Editor/EditorCore/MeshAndElementSelection.cs
+++ b/Editor/EditorCore/MeshAndElementSelection.cs
@@ -173,12 +173,31 @@ namespace UnityEditor.ProBuilder
             var shared = mesh.selectedSharedVertices;
 
             var groups = new List<List<int>>();
-
+            var groupIdx = -1;
+            var i = -1;
             foreach (var index in shared)
             {
                 var coincident = new List<int>();
                 mesh.GetCoincidentVertices(mesh.sharedVerticesInternal[index][0], coincident);
                 groups.Add(coincident);
+                i++;
+                // Make sure the last selected vertic is the last in the group
+                var idx = coincident.IndexOf(mesh.selectedIndexesInternal[mesh.selectedIndexesInternal.Length - 1]);
+                if (idx != -1 && idx < coincident.Count)
+                {
+                    var item = coincident[idx];
+                    groupIdx = i;
+                    coincident[idx] = coincident[coincident.Count - 1];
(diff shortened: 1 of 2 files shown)


Editor/EditorCore/MeshAndElementSelection.cs
```diff
@@ -132,13 +132,34 @@ static IEnumerable<List<Face>> GetFaceSelectionGroups(ProBuilderMesh mesh)
...
 
             foreach (var wing in wings)
             {
                 var group = new List<Face>() {};
                 CollectAdjacentFaces(wing, filter, group);
                 if (group.Count > 0)
+                {
+                    // Make sure the last selected face is the last in the group
+                    var idx = group.IndexOf(mesh.selectedFacesInternal[mesh.selectedFacesInternal.Length - 1]);
+                    if (idx != -1 && idx <= group.Count - 1)
```
neilunity, line 146: idx < group.Count

Editor/EditorCore/MeshAndElementSelection.cs
```diff
@@ -132,13 +132,34 @@ static IEnumerable<List<Face>> GetFaceSelectionGroups(ProBuilderMesh mesh)
...
+                    {
+                        var item = group[idx];
+                        groupIdx = i;
+                        group.RemoveAt(idx);
```
neilunity, line 150: swap last item with this item instead, you save on array operations

Editor/EditorCore/MeshAndElementSelection.cs
```diff
@@ -132,13 +132,34 @@ static IEnumerable<List<Face>> GetFaceSelectionGroups(ProBuilderMesh mesh)
...
+                    }
                     groups.Add(group);
+                    i++;
+                }
+            }
+            // Make sure the last selected face's group is the last in the groups
+            if (groupIdx != -1 && groupIdx <= groups.Count - 1)
+            {
+                var item = groups[groupIdx];
+                groups.RemoveAt(groupIdx);
```
neilunity, line 161: as above, swap

PR #298, opened 2020-06-01
Feature preview pathfinding

RoxanneCor:
### Purpose of this PR

Added a preview for the pathfinding functionality on faces. When holding shift+ctrl, faces that would be selected should be highlighted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Features
 
 - Added iterative selection on edges.
+- Added pathfinding and its preview for faces
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Features
 
 - Added iterative selection on edges.
-- Added pathfinding and its preview for faces
+- Added a selection preview when using the Select Path tool
 
 ### Bug Fixes
 
--- a/Editor/EditorCore/EditorMeshHandles.cs
+++ b/Editor/EditorCore/EditorMeshHandles.cs
@@ -227,7 +227,7 @@ namespace UnityEditor.ProBuilder
             CompareFunction.Always
         };
 #endif
-
+       
         public static void DrawSceneSelection(SceneSelection selection)
         {
             var mesh = selection.mesh;
@@ -238,35 +238,36 @@ namespace UnityEditor.ProBuilder
             var positions = mesh.positionsInternal;
 
             // Draw nearest edge
-            if (selection.face != null)
+            var faces = selection.faces;
+            foreach (var face in faces)
             {
                 using (new TriangleDrawingScope(s_PreselectionColor))
                 {
                     GL.MultMatrix(mesh.transform.localToWorldMatrix);
 
-                    var face = selection.face;
+
                     var ind = face.indexes;
 
                     for (int i = 0, c = ind.Count; i < c; i += 3)
                     {
                         GL.Vertex(positions[ind[i]]);
-                        GL.Vertex(positions[ind[i+1]]);
-                        GL.Vertex(positions[ind[i+2]]);
+                        GL.Vertex(positions[ind[i + 1]]);
+                        GL.Vertex(positions[ind[i + 2]]);
                     }
                 }
             }
-            else if (selection.edge != Edge.Empty)
+            foreach (var edge in selection.edges)
             {
                 using (var drawingScope = new LineDrawingScope(s_PreselectionColor, mesh.transform.localToWorldMatrix, -1f, CompareFunction.Always))
                 {
-                    drawingScope.DrawLine(positions[selection.edge.a], positions[selection.edge.b]);
+                    drawingScope.DrawLine(positions[edge.a], positions[edge.b]);
                 }
             }
-            else if (selection.vertex > -1)
+            foreach (var vertex in selection.vertexes)
             {
                 using (var drawingScope = new PointDrawingScope(s_PreselectionColor, CompareFunction.Always) { matrix = mesh.transform.localToWorldMatrix })
                 {
-                    drawingScope.Draw(positions[selection.vertex]);
+                    drawingScope.Draw(positions[vertex]);
                 }
             }
         }
--- a/Editor/EditorCore/EditorMeshHandles.cs
+++ b/Editor/EditorCore/EditorMeshHandles.cs
@@ -238,14 +238,11 @@ namespace UnityEditor.ProBuilder
             var positions = mesh.positionsInternal;
 
             // Draw nearest edge
-            var faces = selection.faces;
-            foreach (var face in faces)
+            foreach (var face in selection.faces)
             {
                 using (new TriangleDrawingScope(s_PreselectionColor))
                 {
                     GL.MultMatrix(mesh.transform.localToWorldMatrix);
-
-
                     var ind = face.indexes;
 
                     for (int i = 0, c = ind.Count; i < c; i += 3)
--- a/Editor/EditorCore/EditorMeshHandles.cs
+++ b/Editor/EditorCore/EditorMeshHandles.cs
@@ -238,9 +238,9 @@ namespace UnityEditor.ProBuilder
             var positions = mesh.positionsInternal;
 
             // Draw nearest edge
-            foreach (var face in selection.faces)
+            using (new TriangleDrawingScope(s_PreselectionColor))
             {
-                using (new TriangleDrawingScope(s_PreselectionColor))
+                foreach (var face in selection.faces)
                 {
                     GL.MultMatrix(mesh.transform.localToWorldMatrix);
                     var ind = face.indexes;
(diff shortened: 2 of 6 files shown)


Runtime/Core/SelectionPickerSettings.cs
```diff
@@ -1,70 +1,101 @@
...
     {
         public GameObject gameObject;
         public ProBuilderMesh mesh;
-        public int vertex;
-        public Edge edge;
-        public Face face;
+        public List<int> vertexes;
+        public List<Edge> edges;
+        public List<Face> faces;
+
```
RoxanneCor, line 13: I'm not so sure about this part. I did this to simplify cases when we want to set the list to a single value, but it may be confusing?

neilunity: indeed, maybe a clear function name instead

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Features
 
 - Added iterative selection on edges.
+- Added pathfinding and its preview for faces
```
karl-, line 13: This should be descriptive enough that a user could understand the effect of the change. Sometimes this means adding context to the changelog entry. Consider something like "Added a selection preview when using the Select Path tool" or similar.

Editor/EditorCore/EditorMeshHandles.cs
```diff
@@ -238,35 +238,33 @@ public static void DrawSceneSelection(SceneSelection selection)
             var positions = mesh.positionsInternal;
 
             // Draw nearest edge
-            if (selection.face != null)
+            foreach (var face in selection.faces)
```
karl-, line 241: Suggest moving the `foreach(var face in selection.faces)` loop into the `TriangleDrawingScope`. It saves you from setting up the triangle drawing scope and matrix for each face.

Editor/EditorCore/EditorMeshHandles.cs
```diff
@@ -238,35 +238,33 @@ public static void DrawSceneSelection(SceneSelection selection)
...
                         GL.Vertex(positions[ind[i]]);
-                        GL.Vertex(positions[ind[i+1]]);
-                        GL.Vertex(positions[ind[i+2]]);
+                        GL.Vertex(positions[ind[i + 1]]);
+                        GL.Vertex(positions[ind[i + 2]]);
                     }
                 }
             }
-            else if (selection.edge != Edge.Empty)
+            foreach (var edge in selection.edges)
```
karl-, line 256: Same note, set up the drawing scope then iterate edges.

Editor/EditorCore/EditorMeshHandles.cs
```diff
@@ -238,35 +238,33 @@ public static void DrawSceneSelection(SceneSelection selection)
...
             {
                 using (var drawingScope = new LineDrawingScope(s_PreselectionColor, mesh.transform.localToWorldMatrix, -1f, CompareFunction.Always))
                 {
-                    drawingScope.DrawLine(positions[selection.edge.a], positions[selection.edge.b]);
+                    drawingScope.DrawLine(positions[edge.a], positions[edge.b]);
                 }
             }
-            else if (selection.vertex > -1)
+            foreach (var vertex in selection.vertexes)
```
karl-, line 263: Same note, set up the drawing scope then iterate vertices.

Editor/EditorCore/EditorSceneViewPicker.cs
```diff
@@ -32,6 +32,25 @@ static class EditorSceneViewPicker
...
+                return;
+            var mesh = selection.mesh;
+            var face = selection.faces[0];
+            var activeFace = mesh.GetActiveFace();
+            if (activeFace == null || activeFace == face)
+                return;
+
+            var faces = mesh.facesInternal;
+            var pathFaces = SelectPathFaces.GetPath(mesh, Array.IndexOf<Face>(faces, mesh.GetActiveFace()),
+                                 Array.IndexOf<Face>(faces, face));
```
karl-, line 47: You're calling `mesh.GetACtiveFace()` twice here instead of using `activeFace`.

In the interest of saving on allocations, could we add instead an overload of `GetPath` that accepts a `List<Face>` to append to instead of returning a new collection?

Runtime/Core/SelectionPickerSettings.cs
```diff
@@ -1,70 +1,101 @@
...
         public void CopyTo(SceneSelection dst)
         {
             dst.gameObject = gameObject;
             dst.mesh = mesh;
-            dst.face = face;
-            dst.edge = edge;
-            dst.vertex = vertex;
+            dst.faces = faces;
+            dst.edges = edges;
+            dst.vertexes = vertexes;
```
karl-, line 88: This is creating shallow copies, you want to deep copy these lists.

Runtime/Core/SelectionPickerSettings.cs
```diff
@@ -74,9 +105,9 @@ public bool Equals(SceneSelection other)
             if (ReferenceEquals(this, other)) return true;
             return Equals(gameObject, other.gameObject)
                 && Equals(mesh, other.mesh)
-                && vertex == other.vertex
-                && edge.Equals(other.edge)
-                && Equals(face, other.face);
+                && vertexes == other.vertexes
+                && edges.Equals(other.edges)
+                && Equals(faces, other.faces);
```
karl-, line 110: This is doing a reference comparison for the lists, which probably isn't what you want here. Think of it like comparing the pointer address vs. the contents of the collections.

Runtime/Core/SelectionPickerSettings.cs
```diff
@@ -93,9 +124,9 @@ public override int GetHashCode()
             {
                 int hashCode = (gameObject != null ? gameObject.GetHashCode() : 0);
                 hashCode = (hashCode * 397) ^ (mesh != null ? mesh.GetHashCode() : 0);
-                hashCode = (hashCode * 397) ^ vertex;
-                hashCode = (hashCode * 397) ^ edge.GetHashCode();
-                hashCode = (hashCode * 397) ^ (face != null ? face.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ vertexes.GetHashCode();
+                hashCode = (hashCode * 397) ^ edges.GetHashCode();
```
karl-, line 128: `vertexes` and `edges` both need null checks as well now that they're no longer value types.

Runtime/Core/SelectionPickerSettings.cs
karl-: Agreed, either change these to a `Set{Vertexes, Edges, Faces}` function or remove.

Editor/EditorCore/EditorSceneViewPicker.cs
RoxanneCor: The `GetPath` method returns a list of int but we need to transform that to a list of faces. Plus, the code uses a Stack instead of a List so this would make the code in GetPath a bit more complex. I don't know if it's worth it?

Editor/EditorCore/EditorMeshHandles.cs
```diff
@@ -238,35 +238,35 @@ public static void DrawSceneSelection(SceneSelection selection)
             var positions = mesh.positionsInternal;
 
             // Draw nearest edge
-            if (selection.face != null)
+            using (new TriangleDrawingScope(s_PreselectionColor))
             {
-                using (new TriangleDrawingScope(s_PreselectionColor))
+                foreach (var face in selection.faces)
                 {
                     GL.MultMatrix(mesh.transform.localToWorldMatrix);
```
karl-, line 245: Same note for `GL.MultMatrix`, this only needs to be done once per-object, then all faces should use that same matrix.

Editor/EditorCore/EditorSceneViewPicker.cs
karl-: Ok, if it's not going to be a notice-able improvement that's fine.

Runtime/Core/SelectionPickerSettings.cs
```diff
@@ -1,70 +1,93 @@
...
         public void CopyTo(SceneSelection dst)
         {
             dst.gameObject = gameObject;
             dst.mesh = mesh;
-            dst.face = face;
-            dst.edge = edge;
-            dst.vertex = vertex;
+            dst.faces = faces.ConvertAll(x => x);
+            dst.edges = edges.ConvertAll(x => x);
+            dst.vertexes = vertexes.ConvertAll(x => x);
```
karl-, line 80: You might see a performance improvement here if instead of using linq you just cleared and appended to the existing lists. Ex,

```
dst.faces.Clear(); 
foreach(var x in src.faces) 
    dst.faces.Add(x);
```

Runtime/Core/SelectionPickerSettings.cs
```diff
@@ -74,9 +97,9 @@ public bool Equals(SceneSelection other)
             if (ReferenceEquals(this, other)) return true;
             return Equals(gameObject, other.gameObject)
                 && Equals(mesh, other.mesh)
-                && vertex == other.vertex
-                && edge.Equals(other.edge)
-                && Equals(face, other.face);
+                && Enumerable.SequenceEqual(vertexes, other.vertexes)
+                && Enumerable.SequenceEqual(edges, other.edges)
+                && Enumerable.SequenceEqual(faces, other.faces);
```
karl-, line 102: General question, how expensive is `SequenceEqual`? Is it notice-able in the profiler when you're moving around the scene?

RoxanneCor: Just checked with the profiler and the operation does no GC Alloc and takes about 0.01ms. So it doesn't seem heavy at all.

PR #299, opened 2020-06-04
Bugfix/ctrl selection

lopezt-unity:

### Purpose of this PR
Bug fix for selection with dragging while holding ctrl key

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1252394/

### Comments to Reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [UNRELEASED]
 
 ### Bug Fixes
-- [case: 1252394] Fixed dragging selection with Ctrl problem.
+- [case: 1252394] Fixed dragging selection with Ctrl problem (and key pressed problem in general).
 
 ## [4.4.0-preview.1] - 2020-06-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [UNRELEASED]
+
+### Bug Fixes
+- [case: 1252394] Fixed dragging selection with Ctrl problem.
+
 ## [4.4.0-preview.1] - 2020-06-29
 
 ### Features
--- a/Editor/EditorCore/ProBuilderEditor.cs
+++ b/Editor/EditorCore/ProBuilderEditor.cs
@@ -590,6 +590,10 @@ namespace UnityEditor.ProBuilder
                 if (m_CurrentEvent.keyCode == KeyCode.Escape && selectMode != SelectMode.Object)
                 {
                     selectMode = SelectMode.Object;
+
+                    m_IsDragging = false;
+                    m_IsReadyForMouseDrag = false;
+                    
                     m_CurrentEvent.Use();
                 }
             }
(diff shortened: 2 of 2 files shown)


karl-:
Good catch Joel. We should probably look at moving the selection code out of `ProBuilderEditor` and use the `GUIUtility.hotControl` properly.

lopezt-unity:
Indeed it seems a little more complicated than that! I’ll go back on this 😊 Thomas Sent from Mail for Windows 10 From: Karl HenkelSent: Thursday, June 4, 2020 7:57 PMTo: Unity-Technologies/com.unity.probuilderCc: Thomas Lopez; AuthorSubject: Re: [Unity-Technologies/com.unity.probuilder] Bugfix/ctrl selection (#299) Good catch Joel. We should probably look at moving the selection code out of ProBuilderEditor and use the GUIUtility.hotControl properly.—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.

lopezt-unity:
Hi Joel, Karl, So here is where I am now : not only the problem appears with the trl key but with all the keys you press before starting a selection with a dragging in sub-scene mode (Space bar or A-key for instance).It seems to me that this bug was introduced by the ew Unity Shortcut system as previously these keys where I think filtered with : if (m_CurrentEvent.type == EventType.KeyDown){    if (s_Shortcuts.value.Any(x => x.Matches(m_CurrentEvent.keyCode, m_CurrentEvent.modifiers)))        m_CurrentEvent.Use();}which is not used anymore in unity versions since 2019.1. To better correct this bug I need to understand better the primary idea behind this :if (EditorHandleUtility.SceneViewInUse(m_CurrentEvent) || m_CurrentEvent.isKey && m_IsDragging)If I’m right the idea is to interrupt the dragging if either a scene view manipulation is detected or a key is pressed after a dragging has been started. The problem coming from the fact that the keydown event is spamming repeatedly the system and not only on the frame where it is pressed. If this is the desired behavior, I could just check if the key has been pressed for the first time on this frame and it should work (I have an idea to make this work). Otherwise I found a way for the sceneview events to stop the dragging but the other keys doesn’t stop it. This can lead to some undesired artifacts, for instance the ‘F’ focuses on the selected vertex but the dragging is not stopped. I don’t see this as an acceptable behavior then… Can I have your inputs on this ?Thanks! Thom   Thomas LopezTools Developerthomas.lopez@unity3d.comunity.com  From: Thomas LopezSent: Friday, June 5, 2020 9:46 AMTo: Unity-Technologies/com.unity.probuilder; Unity-Technologies/com.unity.probuilderCc: AuthorSubject: RE: [Unity-Technologies/com.unity.probuilder] Bugfix/ctrl selection (#299) Indeed it seems a little more complicated than that! I’ll go back on this 😊 Thomas Sent from Mail for Windows 10 From: Karl HenkelSent: Thursday, June 4, 2020 7:57 PMTo: Unity-Technologies/com.unity.probuilderCc: Thomas Lopez; AuthorSubject: Re: [Unity-Technologies/com.unity.probuilder] Bugfix/ctrl selection (#299) Good catch Joel. We should probably look at moving the selection code out of ProBuilderEditor and use the GUIUtility.hotControl properly.—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.

lopezt-unity:
It should be ok right now @JoelFortin, let me know if not.
We were discussing with Karl also to make it another (and cleaner) way. I will look at this next week I think.

lopezt-unity:
I just fixed the problem with pressing ESC you just mentioned. Indeed it was not really great to constantly have the rect selection following you...

JoelFortin:
Tested with the ESC key fix ... look good, thanks.
But we need to consider adding an Undo support for when we loose the selection like that so we at least get our selection back otherwise it's destructive. This problem seems to be generalized so it's not something we should fix in this PR.

PR #303, opened 2020-06-10
Shape/draw tool

RoxanneCor:
### Purpose of this PR

Draw Shape Tool

### Comments to Reviewers

A big part of the code is Karl's :)

For now this only do the switching shape and parameters from the inspector. The rotation, snapping to grid and UX will come in another PR.

Also, there are lots of new classes to represent the shapes. A refactor will be made along with the UX to make the Build Shape Tool use those shapes as well to avoid duplicate code.

(This is only a draft, not gonna push into master before it's completed)


RoxanneCor:
I fixed the 2 last errors, but I can't reproduce the first one. Is it still happening ? (maybe it got fixed along with the rest)

PR #306, opened 2020-06-16
Shape/draw tool rotation

RoxanneCor:
### Purpose of this PR

Added the rotation feature to the Draw Shape Tool

### Comments to Reviewers

We will probably need to change the UX of this feature, but the functionality should be complete.
When you click on a handle of one of the visible face of the bounding box around the shape, the shape should rotate accordingly.


Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -184,6 +184,20 @@ void SelectPlane(Event evt)
             }
         }
 
+        Vector3 ToEularAngles(Vector3 diff)
```
neilunity, line 187: Euler
this does not convert to euler, consider renaming

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -196,7 +210,12 @@ void SetWidthAndDepth(Event evt)
                     if (m_Plane.Raycast(ray, out distance))
                     {
                         m_OppositeCorner = ray.GetPoint(distance);
+                        var diff = m_OppositeCorner - m_Origin;
                         m_HeightCorner = m_OppositeCorner;
+
+                        if (m_Shape != null)
```
neilunity, line 216: does the rebuild and repaint also depend on m_Shape not being null ?

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -223,6 +242,12 @@ void SetHeight(Event evt)
                 {
                     Ray ray = HandleUtility.GUIPointToWorldRay(evt.mousePosition);
                     m_HeightCorner = Math.GetNearestPointRayRay(m_OppositeCorner, m_Plane.normal, ray.origin, ray.direction);
+
+                    var diff = m_HeightCorner - m_Origin;
+
+                    if (m_Shape != null)
```
neilunity, line 248: same as above

PR #307, opened 2020-06-26
Ensure "ProBuilderize" is available at window init

aybe:
### Purpose of this PR

Currently, if you have an active selection in Hierarchy that hasn't been ProBuilder-ized yet and open the main window, the action is not enabled.

This PR enables the action immediately, instead of having to deselect the object and select it again so the action is finally enabled in the action palette.

### Comments to Reviewers

That's my first contribution, I hope I didn't forget anything and it's properly done, let me know otherwise !

### DIFF
--- a/Editor/EditorCore/EditorToolbar.cs
+++ b/Editor/EditorCore/EditorToolbar.cs
@@ -67,6 +67,8 @@ namespace UnityEditor.ProBuilder
             isIconMode = ProBuilderEditor.s_IsIconGui;
             this.window = ProBuilderEditor.instance;
             CalculateMaxIconSize();
+
+            MeshSelection.ClearElementSelection();
         }
 
         void OnDisable()
--- a/Editor/EditorCore/EditorToolbar.cs
+++ b/Editor/EditorCore/EditorToolbar.cs
@@ -67,8 +67,6 @@ namespace UnityEditor.ProBuilder
             isIconMode = ProBuilderEditor.s_IsIconGui;
             this.window = ProBuilderEditor.instance;
             CalculateMaxIconSize();
-
-            MeshSelection.ClearElementSelection();
         }
 
         void OnDisable()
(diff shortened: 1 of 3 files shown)


Editor/EditorCore/EditorToolbar.cs
```diff
@@ -67,6 +67,8 @@ void OnEnable()
             isIconMode = ProBuilderEditor.s_IsIconGui;
             this.window = ProBuilderEditor.instance;
             CalculateMaxIconSize();
+
+            MeshSelection.ClearElementSelection();
```
karl-, line 71: Without checking, it looks like this would clear the element selection on domain reloads. Could this be resolved instead by modifying the constructor of the ProBuilderize menu action? Looking at the `ProBuilderize()` constructor it looks like it doesn't actually set `m_Enabled` when it's instantiated.

aybe:
Wrong button pressed, I was trying to figure out what Resolve conversation does but it vanished.

Yes, makes perfect sense, I've reverted the change and did as you said, simpler and better !

aybe:
I would like to ask you about something else in here since there isn't any Issues tab in this project where I could ask a question in.

I was thinking of implementing 'live update' face selection, that is, that ProBuilder shows actual selection while still dragging.

By calling `DoMouseDrag` during `EventType.MouseDrag` it does work though now there's an unfortunate situation when you want to <kbd>Shift</kbd> faces to the actual selection; the newest selection gets inverted.

I suppose the initial selection should be kept in memory somehow, however, the code is pretty difficult to understand to be honest.

How would you go about updating selection while the user is dragging, knowing that there can already be a selection and knowing that there are different selection modes ?

Thank you.

karl-:
> I was thinking of implementing 'live update' face selection, that is, that ProBuilder shows actual selection while still dragging.

There is actually a branch here https://github.com/Unity-Technologies/com.unity.probuilder/pull/298 that does this for the new path selection feature. I don't think it would be too difficult to adapt to marquee selection. In fact, I suspect the blocker for this will be picking performance.

aybe:
Great, I will look at it then.

Thank you!

PR #309, opened 2020-06-29
Highlight occluded element selection with a tint

karl-:

### Purpose of this PR

Render selected mesh elements that are occluded with a slightly darker tint. This replaces the old method of just not rendering anything.

face
edge
vert


lopezt-unity:
Looks ok to me :)
It seems that Render method on line 323 in EditorMeshHandles.cs is not used anymore (according to Rider :) ).
If it stays there, should that also handles color with occlusion tint ? 
That's my only remark :)

karl-:
@lopezt-unity removed unused function cb4fc0a

karl-:
Do we have a ticket for the `m_WireMaterial` null reference still? That should be done in a separate case, but definitely needs to be high priority.

I like the idea of making this optional. Do you see that as living in the Preferences window, or does it need to be a toolbar option? @gabrielw-us would like your thoughts on this as well 👆

JoelFortin:
The ticket for the null reference is there -->
https://fogbugz.unity3d.com/f/cases/1251289/

I have no problem having a preference setting to decide what's my default behavior but I'd also want to map this to a shortcut to quickly toggle the visual when needed. Or maybe a "See-Through" item in the PB window that show ON or OFF like the "Select Hidden" option for instance.

karl-:
@gabrielw-us @JoelFortin to provide

- [ ] Name of the action
- [ ] Icon for toolbar
- [ ] Default shortcut keys

karl-:
@JoelFortin @gabrielw-us cfa436e adds the menu item and shortcut. I called the action "X-Ray" and gave it a default shortcut of `Alt + Shift + X`, feel free to propose alternatives.

It also still needs an icon for the toolbar.

karl-:
- [ ] documentation

gabrielw-us (approved):

PR #311, opened 2020-07-07
fix vertex colors not applying linear values in linear color space

karl-:
### Purpose of this PR

Fixes vertex colors applying gamma corrected values when color space is linear.

https://forum.unity.com/threads/probuilder-applying-incorrect-vertex-colors-when-using-linear-colorspace.927107/#post-6064661


gabrielw-us:
Ah, THIS is the reason my linear-space VR game has been looking so wonky! :D I was too lazy to try and repro ... but had a suspicion. Glad this other fellow brought it up properly, and thanks for the quick fix!

PR #312, opened 2020-07-08
Fixed issue where the dimension overlay visibility would reset

antoine-unity:
### Purpose of this PR

Fix the issue where the dimension overlay visibility would reset when switching to play mode or reopening the editor.

### Links

**Fogbugz:** [case#1259845](https://fogbugz.unity3d.com/f/cases/1259845/)


karl-:
Might want to edit the changelog to note that this fix is only applicable 2019.1+

PR #314, opened 2020-07-09
Fixed Shortcuts Not Being Saved in 2018.4

antoine-unity:
### Purpose of this PR

Preferences are normally change when the value is set but in the case of shortcut we modify the array directly. This PR ensure that changing the shortcut from the UI saves the changes.

### Links

**Fogbugz:** [case#1259506](https://fogbugz.unity3d.com/f/cases/1259506/)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Bug Fixes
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [case: 1249056] Fixed Ctrl selection being inconsistent.
 - [case: 1247270] Fixed `Null Reference Exception` when entering Prefab Stage after merging multiple `ProBuilderMesh`.
 - [case: 1252394] Fixed dragging selection with Ctrl problem (and key pressed problem in general).
+- [case: 1259506] Fixed shortcut not being saved in 2018.4.
```
karl-, line 29: 4.4.0-preview.1 has already shipped, this needs to be under `### [Unreleased]`

PR #321, opened 2020-08-03
bugfix for case 1262236 - verifying that the mesh to export as at lea…

lopezt-unity:
When exporting a mesh, verify that it contains at least a model with vertices to avoid exporting empty meshes.
Empty mesh was creating a null ref exception when trying to export an empty polyshape.

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1262236/

### Comments to Reviewers


Editor/EditorCore/ObjExporter.cs
```diff
@@ -96,6 +96,15 @@ public static bool Export(string name, IEnumerable<Model> models, out string obj
                 return false;
             }
 
+            if (models.Where(m => m.vertices.Any()).Count() < 1)
```
karl-, line 99: Could use `m.vertexCount` here to avoid creating copies of the vertex array.

lopezt-unity: I've just updated this 👍

PR #324, opened 2020-08-06
Bugfix/1256154 stackoverflow on probuilderizes

lopezt-unity:
### Purpose of this PR

When Probuilderizing a dense-flat mesh, the recursive call to FindSoftEdgesRecursive was causing a StackOverflow Exception as the exploration was too deep regarding the number of edges of the mesh. 
=> The exception is caught and a message is shown to the user to let him know that probuilderize action is done but not smoothing

Edges of Probuilderized meshes were not correctly displayed if edges count > ushort.maxValue. 
This has been fixed here.
Old:
image

New:
image

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1256154/

### Comments to Reviewers


lopezt-unity:
I did try to run the code, it didn't seem to be cause by the geometry itself as it was not bouncing back and forth between the same geometry elements. But we can definitively have a deeper look at this indeed!

PR #325, opened 2020-08-07
Fixed PolyShape in Prefab mode

lopezt-unity:
### Purpose of this PR
PolyShape fixes:
- PolyShapeMode was not serialized when exiting prefab mode, exiting prefab mode after building a polyshape was nor working.
- Update point insertion visualization, repaint was not called enough when moving the mouse around the shape.

### Comments to Reviewers


Editor/EditorCore/PolyShapeEditor.cs
```diff
@@ -426,11 +432,18 @@ void DoPointPlacement()
...
                     if (distanceToVertex > PreferenceKeys.k_MaxPointDistanceFromControl && distanceToLine < PreferenceKeys.k_MaxPointDistanceFromControl)
+
+                    if(distanceToVertex > PreferenceKeys.k_MaxPointDistanceFromControl &&
+                       distanceToLine < PreferenceKeys.k_MaxPointDistanceFromControl)
                     {
-                        Handles.color = Color.green;
-                        Handles.DotHandleCap(-1, wp, Quaternion.identity, HandleUtility.GetHandleSize(wp) * k_HandleSize, evt.type);
+                        if(evt.type == EventType.Repaint)
+                        {
+                            Handles.color = Color.green;
```
antoine-unity, line 441: Normally, I would say to use a Handles.DrawingScope for color changes but the value seems to be reset to white at the end so it should be fine

Editor/EditorCore/PolyShapeEditor.cs
```diff
@@ -444,6 +457,10 @@ void DoPointPlacement()
 
                         Handles.color = Color.white;
                     }
+
+                    if(evt.type != EventType.Repaint)
```
antoine-unity, line 461: If you don't want to trigger a SceneView.RepaintAll each frame, I would include EventType.Layout in there. IMGUI "frame" are always composed of Layout+Repaint+other events. Otherwise this could be switched to only repaint all on layout.

PR #327, opened 2020-08-10
Point To Point Cut Tool

lopezt-unity:
### Purpose of this PR

New feature: Adding the point-to-point cut tool.
Using this tool the user can add a cut in the face of a PB Mesh, creating a division of the original face in different sub-faces.
cutFinal

This tool is based on `EditorTool`, so it is made to be used with Unity 2019.4 and above.

### Comments to Reviewers


gabrielw-us:
Awesome 👍 

I noticed that edges do not highlight unless their face is visible, is this on purpose? Feels unexpected to me.

gabrielw-us:
Also a bit unsure about the ESC key action, and "commit" result when path is invalid. Both times, very easy to lose complex cut, and no way to retrieve it, no warning this will happen. Also, works differently than Polyshape. I'll investigate! :)

JoelFortin (approved):
Peer-review with Thomas ... all the problems found in the first pass was fixed and looks pretty solid to me right now. Nice Work !!!

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+                m_Types = types;
+            }
+        }
+
+        static readonly Color k_HandleColor = new Color(.8f, .8f, .8f, 1f);
+        static readonly Color k_HandleColorAddNewVertex = new Color(.01f, .9f, .3f, 1f);
+        static Color s_HandleColorAddVertexOnEdge = new Color(.3f, .01f, .9f, 1f);
+        static Color s_HandleColorUseExistingVertex = new Color(.01f, .5f, 1f, 1f);
+        static readonly Color k_HandleColorModifyVertex = new Color(1f, .75f, .0f, 1f);
+        const float k_HandleSize = .05f;
```
karl-, line 76: should we reuse any of the existing `Handles` pref colors here? cc @gabrielw-us

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+        bool m_ModifyingPoint; // State machine instead? //Status
+        int m_SelectedIndex = -2;
+        bool m_IsCutValid = true;
+
+        //Cut tool elements
+        internal Face m_TargetFace;
+        internal Face m_CurrentFace;
+        internal Vector3 m_CurrentPosition = Vector3.positiveInfinity;
+        internal Vector3 m_CurrentPositionNormal = Vector3.up;
+        internal VertexTypes m_CurrentVertexTypes = VertexTypes.None;
```
karl-, line 98: why are these all internal? who is accessing these outside of the class?

Editor/EditorCore/CutTool.cs
karl-: oh I see, they do later on in the ctor

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+        {
+            None = 0 << 0,
+            NewVertex = 1 << 0,
+            AddedOnEdge = 1 << 1,
+            ExistingVertex = 1 << 2,
+            VertexInShape = 1 << 3,
+        }
+
+        [Serializable]
+        internal struct CutVertexData
```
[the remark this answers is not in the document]
karl-: would it be so that this tool continues to function correctly if a domain reload occurs during operation?

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+        /// </summary>
+        private void UndoRedoPerformed()
+        {
+            if(m_CutPath.Count == 0)
+                m_TargetFace = null;
+
+            m_SelectedIndex = -1;
+            m_MeshConnections.Clear();
+
+            EditorApplication.delayCall = () => RebuildCutShape();
```
karl-, line 191: why is this done as a delayed call?

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+            EditorGUI.indentLevel--;
+
+            GUI.enabled = true;
+
+            if(MeshSelection.selectedObjectCount != 1)
+                GUI.enabled = false;
+
+            if(m_Mesh == null)
+            {
+                if(GUILayout.Button("Start"))
```
karl-, line 278: use `EditorGUIUtility.TrContent` for translation

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+        {
+            m_MeshConnections.Clear();
+            if(m_CutPath.Count < 2)
+                return;
+
+            List<Vector3> existingVerticesInCut =
+                m_CutPath.Where(v => ( v.types | VertexTypes.ExistingVertex ) != 0)
+                         .Select(v => v.position).ToList();
+
+            Vertex[] vertices = m_Mesh.GetVertices();
```
karl-, line 488: if you're not adding or removing vertices you can save an alloc here by using `m_Mesh.positionsInternal`

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+                    RebuildCutShape();
+                    evt.Use();
+                    break;
+                }
+
+                case KeyCode.Escape:
+                    evt.Use();
+                    Clear();
+                    //Leave the current tool
+                    Tools.current = Tool.None;
```
karl-, line 1172: Use `EditorToolManager.RestorePreviousTool();`

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+        /// Visual indications to help the user: highlighting faces, edges and vertices when snapping on them
+        /// </summary>
+        void DoVisualCues()
+        {
+            if(m_Mesh != null)
+            {
+                if(m_TargetFace == null)
+                {
+                    EditorHandleDrawing.HighlightVertices(m_Mesh, m_Mesh.sharedVertexLookup.Keys.ToArray(), false);
+                    EditorHandleDrawing.HighlightEdges(m_Mesh, m_Mesh.faces.SelectMany(f => f.edges).Distinct().ToArray(),
```
karl-, line 1247: it would be nice to avoid the `ToArray` and linq calls here if we're running this on `EventType.Repaint`

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+                        false);
+
+                    if(m_CurrentFace != null)
+                        EditorHandleDrawing.HighlightFaces(m_Mesh, new Face[]{m_CurrentFace}, Color.Lerp(Color.blue, Color.cyan, 0.5f));
+                }
+                else
+                {
+                    var edges = m_TargetFace.edges;
+                    EditorHandleDrawing.HighlightVertices(m_Mesh, edges.Select(e => e.a).ToArray(), false);
+                    EditorHandleDrawing.HighlightEdges(m_Mesh, edges.ToArray(), false);
```
karl-, line 1257: same note as above, can we reduce the amount of allocations by adding the appropriate overloads to `EditorHandleDrawing`?

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+
+            // While the vertex count may not change, the triangle winding might. So unfortunately we can't take
+            // advantage of the `vertexCountChanged = false` optimization here.
+            ProBuilderEditor.Refresh();
+            SceneView.RepaintAll();
+        }
+
+        void ValidateCutShape()
+        {
+            Vertex[] vertices = m_Mesh.GetVertices();
```
karl-, line 1294: if this function is just reading the vertex data use `m_Mesh.positionsInternal` to skip the array copy.

Editor/EditorCore/PolyShapeEditor.cs
```diff
@@ -750,10 +750,10 @@ void OnFinishVertexMovement()
...
         void UndoRedoPerformed()
         {
-            if (polygon.polyEditMode == PolyShape.PolyEditMode.None)
-                ProBuilderEditor.selectMode = ProBuilderEditor.selectMode & ~(SelectMode.InputTool);
-            else
-                ProBuilderEditor.selectMode = ProBuilderEditor.selectMode | SelectMode.InputTool;
+            // if (polygon.polyEditMode == PolyShape.PolyEditMode.None)
+            //     ProBuilderEditor.selectMode = ProBuilderEditor.selectMode & ~(SelectMode.InputTool);
+            // else
+            //     ProBuilderEditor.selectMode = ProBuilderEditor.selectMode | SelectMode.InputTool;
```
karl-, line 756: what's going on here?

Runtime/Core/SubdivideOnVertex.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+
+namespace UnityEngine.ProBuilder
+{
+    [DisallowMultipleComponent]
+    [RequireComponent(typeof(ProBuilderMesh))]
+    [ExcludeFromPreset, ExcludeFromObjectFactory]
+    public sealed class SubdivideOnVertex : MonoBehaviour
```
karl-, line 12: let's talk about this approach, not sure about using a monobehaviour here

karl- (changes requested):

karl-:
Screen Shot 2020-10-07 at 4 21 28 PM (2)

on macos the cursor size is incorrect

Editor/EditorCore/CutTool.cs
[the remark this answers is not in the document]
lopezt-unity: When it's not serializable the undo does not seem to work properly as positions from the cut are not serialized. Then Undo  does not remove the last position from the cut if the structure is not serializable.

Editor/EditorCore/CutTool.cs
```diff
@@ -0,0 +1,1421 @@
...
+
+            public Vector3 position
+            {
+                get => m_Position;
+                set => m_Position = value;
+            }
+
+            public Vector3 normal
+            {
+                get => m_Normal;
```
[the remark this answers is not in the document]
lopezt-unity: only getter is needed, same for the type, the set are only done using the constructor.
This can be modified if needed.

Editor/EditorCore/CutTool.cs
lopezt-unity: The test class CutToolTest.cs is :) (Yeah, I did try to make a test class for this tool :D )

Editor/EditorCore/CutTool.cs
lopezt-unity: I think it was for a purpose that is not valid anymore => I removed it

Editor/EditorCore/CutTool.cs
lopezt-unity: What is the difference between this method and L10n.Tr method?

Runtime/Core/SubdivideOnVertex.cs
lopezt-unity: Oops, this was only a test functionality I did when starting in the team, I'm removing this class

Editor/EditorCore/PolyShapeEditor.cs
lopezt-unity: > on macos the cursor size is incorrect
> 
> also to note - the cursor images are 512^2, which is way larger than they need to be in the first place. importing icons is the main cause for lengthy package import times.

The cursor images are now 64x64, feel free to modify the aspect on MacOS, I am on Windows and don't really know how to handle that in Mac :p

lopezt-unity: This is not valid anymore after the merge with the new polyshape as editortool

lopezt-unity:
> Screen Shot 2020-10-07 at 4 21 28 PM (2)
> 
> on macos the cursor size is incorrect
> 
> also to note - the cursor images are 512^2, which is way larger than they need to be in the first place. importing icons is the main cause for lengthy package import times.

Icons are now 64x64 pixels, which should be better.
@karl- could you correct the icon display size problem on MacOS, as I don't have one?

Editor/EditorCore/PolyShapeTool.cs
```diff
@@ -761,14 +761,13 @@ void HandleKeyEvent(Event evt)
                     {
                         DestroyImmediate(polygon.gameObject);
                         DestroyImmediate(this);
+                        ProBuilderEditor.ResetToLastSelectMode();
```
karl-, line 764: This might not be necessary anymore, I'm not sure.

Runtime/Core/ProBuilderMesh.cs
```diff
@@ -466,6 +466,69 @@ public Vertex[] GetVertices(IList<int> indexes = null)
             return v;
         }
 
+                /// <summary>
+        /// Creates a new array of vertices with values from a @"UnityEngine.ProBuilder.ProBuilderMesh" component.
+        /// </summary>
+        /// <param name="indexes">An optional list of indexes pointing to the mesh attribute indexes to include in the returned Vertex array.</param>
+        /// <returns>An array of vertices.</returns>
```
karl-, line 473: xml docs don't match the function signature

Runtime/MeshOperations/QuadUtility.cs
```diff
@@ -0,0 +1,151 @@
+﻿
+using System.Collections.Generic;
+
+namespace UnityEngine.ProBuilder.MeshOperations
+{
+    public static class QuadUtility
```
karl-, line 6: nice! it's good to see this abstracted

Tests/Editor/Geometry/CutToolTest.cs
```diff
@@ -0,0 +1,189 @@
...
+    SelectMode m_PreviousSelectMode;
+
+    [SetUp]
+    public void Setup()
+    {
+        // make sure the ProBuilder window is open
+        if (ProBuilderEditor.instance == null)
+            ProBuilderEditor.MenuOpenWindow();
+
+        Assert.That(ProBuilderEditor.instance, Is.Not.Null);
```
karl-, line 29: Can use `Assume` when doing test state setup

karl- (approved):

Editor/EditorCore/PolyShapeTool.cs
lopezt-unity: This is not needed for the tool itself (move/rotate/scale) but is indeed needed to retrieve the previous PB mode (face/edge/vertex)

lopezt-unity:
@karl- Could you have a look to the cursor icon issue on iOS by any chance?

karl-:
> @karl- Could you have a look to the cursor icon issue on iOS by any chance?

@lopezt-unity I haven't yet. I don't want to block the PR for that though, let's just merge this and open a bug to fix it

PR #328, opened 2020-08-11
Editortool/polyshape

lopezt-unity:
### Purpose of this PR

PolyShape tool upgrade as `EditorTool`. 
The new version offers the same functionalities than the previous version but is based on the new EditorTools that are part of Unity API.
Small upgrades :
- Menu button is now a toggle to enter/leave polyshape edition
- An overlay window display information about the edition of the polyshape in use
- Custom cursors added when editing the polyshape (ArrowPlus to add position and MoveArrow to move points)
- New version of the point insertion to have a better correspondance between cursor position and point in the polyshape.

### Links

**Jira:** Part of : https://jira.unity3d.com/browse/STO-1476

### Comments to Reviewers


karl-:
General notes (some aren't specific to tools work and are just issues with the Poly Shape)

- [ ] `Extrusion` and `Flip Normals` properties are no longer shown in the Inspector.
- [ ] Pressing `Escape` clears the entire shape regardless of editing state. I don't think that the `Escape` key should ever clear the data. There was an exception at one point for when a Poly Shape is first created, where if insufficient points were provided in the first placement it would destroy the object. However I think even that is probably not desirable.
- [ ] Occasionally the point placement is completely wrong - I haven't found a pattern to this yet.
- [ ] The "Quit Editing" button is not necessary
- [ ] `PolyShapeTool` should be a component tool with `[EditorTool("Edit PolyShape", typeof(PolyShape))]`
- [ ] Need an icon for the tool (pinging @gabrielw-us)

The menu item highlight looks good, that's a nice touch 👍

Editor/EditorCore/MenuAction.cs
```diff
@@ -246,11 +248,31 @@ protected virtual MenuActionState optionsMenuState
...
+        /// <summary>
+        /// Perform whatever action this menu item is supposed to do. You are responsible for implementing Undo.
+        /// </summary>
+        /// <returns>A new ActionResult with a summary of the state of the action's success.</returns>
+        public ActionResult DoAction()
+        {
+            actionChanged(this);
+            return DoAction_Internal();
+        }
+
```
karl-, line 270: I have a vague memory of discussing this, but can you remind me why it is necessary to add this event? Additionally, this will need to go through the a deprecation phase before changing the signature.

Editor/EditorCore/MenuToggle.cs
```diff
@@ -0,0 +1,198 @@
...
+    {
+        public enum MenuToggleState
+        {
+            Inactive,
+            Active,
+        };
+
+        MenuToggleState m_CurrentState;
+
+        protected delegate void StartEndCallBack();
```
karl-, line 27: Use Action<> for delegates https://ono.unity3d.com/unity-extra/unity-meta/raw/@/ReferenceSource/CSharp/Assets/CSharpReference.cs

Editor/EditorCore/MenuToggle.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        public override void OnActionChanged(MenuAction action)
+        {
+            if(m_CurrentState == MenuToggleState.Active && !this.Equals(action))
+            {
+                Debug.Log("Changing menu action! Deactivating current toggle");
+                EndActivation(OnEnd);
+            }
+        }
```
karl-, line 45: A general thought on `MenuToggleState`, could this be handled instead by defining an associated EditorTool and checking that it is active?

Editor/EditorCore/MenuToggle.cs
```diff
@@ -0,0 +1,198 @@
...
+        protected abstract ActionResult StartActivation(StartEndCallBack callback);
+
+        /// <summary>
+        /// Call the Update for the current tool.
+        /// </summary>
+        public void OnUpdate()
+        {
+            if(m_CurrentState == MenuToggleState.Active)
+                UpdateAction();
+        }
```
karl-, line 76: EditorTool manages it's own lifecycle. The Update/End action callbacks seem unnecessary, while Start could be put in the `DoAction` function

Editor/EditorCore/PolyShapeTool.cs
```diff
@@ -0,0 +1,958 @@
...
+#if !UNITY_2020_2_OR_NEWER
+using ToolManager = UnityEditor.EditorTools.EditorTools;
+#else
+using ToolManager = UnityEditor.EditorTools.ToolManager;
+#endif
+
+
+namespace UnityEditor.ProBuilder
+{
+    public class PolyShapeTool : EditorTool
```
karl-, line 22: Missing `[EditorTool("Edit PolyShape", typeof(PolyShape))]`

Editor/EditorCore/PolyShapeTool.cs
```diff
@@ -0,0 +1,958 @@
...
+        void OnFinishVertexMovement()
+        {
+            m_IsModifyingVertices = false;
+            RebuildPolyShapeMesh(polygon);
+        }
+
+        void UndoRedoPerformed()
+        {
+            ClearLineRenderers();
+            InitLineRenderers();
```
karl-, line 952: Now that this is an `EditorTool` is it necessary to `InitLineRenderers` on Undo still?

Editor/EditorCore/PolyShapeTool.cs
```diff
@@ -0,0 +1,958 @@
...
+
+            if (EditorGUI.EndChangeCheck())
+            {
+                if (polygon.polyEditMode == PolyShape.PolyEditMode.None)
+                {
+                    if (ProBuilderEditor.instance != null)
+                        ProBuilderEditor.instance.ClearElementSelection();
+
+                    UndoUtility.RecordObject(polygon, "Change Polygon Shape Settings");
+                    UndoUtility.RecordObject(polygon.mesh, "Change Polygon Shape Settings");
```
karl-, line 300: Could use `UndoUtility.RecordComponents<PolyShape, ProBuilderMesh>(targets as Component[], "Change Polygon Shape Settings");` here

karl-:
Overall it looks good. Aside from the usual style and polish questions, I'd be interested to see if we can avoid modifying `DoAction` and adding the `actionChanged` action.

Editor/EditorCore/MenuAction.cs
lopezt-unity: The event was added because the toggle tools (polyshape, cut tool, shape tool), need to be aware of probuilder menuaction changes so that it can disable a toggle tool when selecting a new PB action. 
If you have another idea to make this, it could be better :)

I was reusing the method DoAction, thinking that might be transparent, but indeed a deprecation phase would be better if this structure is kept.

Editor/EditorCore/PolyShapeTool.cs
lopezt-unity: I didn't add the attribute so that the tool does not appear in the custom tool menu but only in the PB Menu. 
Should this be modified?

Editor/EditorCore/PolyShapeTool.cs
lopezt-unity: I think it is as you can undo "add point to polyshape" thus the polyshape path must be updated accordingly.

Editor/EditorCore/MenuToggle.cs
lopezt-unity: I changed it ! :)

Editor/EditorCore/MenuToggle.cs
lopezt-unity: Agree on the UpdateAction, this does not seems necessary, this was a first version of the menu toggle.
However, the problem with the end is still the same as a user can select another PB action in the menu without ending the tool first. As other actions are not EditorTools, do you have an idea on how to handle this?

lopezt-unity:
Karl, some feedbacks regarding your notes:

- [x]  Extrusion and Flip Normals properties are no longer shown in the Inspector.
_=> I re-added this possibility (not that easy with managing EditorTool states and selection :p )_
- [ ]  Pressing Escape clears the entire shape regardless of editing state. I don't think that the Escape key should ever clear the data. There was an exception at one point for when a Poly Shape is first created, where if insufficient points were provided in the first placement it would destroy the object. However I think even that is probably not desirable.
_=> Could we discuss about the desired behavior on this together ?_
- [ ]  Occasionally the point placement is completely wrong - I haven't found a pattern to this yet.
_=> Never run into that case hopefully, let me know if you run into that again and find the reason_
- [ ]  The "Quit Editing" button is not necessary
_=> This is a button that mirrors the previous behavior of the button "Editing PolyShape" in the precedent version of the tool, should this be removed then, or put this back on the Editor script? but as this impacts the behavior of the tool I prefer to put that in the tool overlay._
- [ ]  PolyShapeTool should be a component tool with [EditorTool("Edit PolyShape", typeof(PolyShape))]
_=> See my remark on this, my first thought was not to add the tool in the Custom tool menu but we can talk about the desired behavior of the tool together_
- [ ]  Need an icon for the tool (pinging @gabrielw-us)
_=> The tool is the same, so we should keep the same icon for the PB menu and as for now the tool is not added to the editor tool menu, I guess it's ok if it remains this way_
- [ ] The menu item highlight looks good, that's a nice touch 👍
_=> Cool :)_

Editor/EditorCore/PolyShapeTool.cs
lopezt-unity: I've done that using polygon instead of targets however

gabrielw-us:
Super cool! Nice work!

gabrielw-us (approved):

gabrielw-us:
Note: I expect we'll work on the visuals a bit, just polish etc, later

lopezt-unity:
> Note: I expect we'll work on the visuals a bit, just polish etc, later

Whenever you want! I'll have time this week

JoelFortin (approved):
Peer-reviewed with Thomas ... LGTM.

karl- (approved):

karl-:
Screen Shot 2020-10-07 at 4 21 28 PM (2)

on macos the cursor image is not the correct size.

PR #330, opened 2020-08-12
Shape/draw tool

RoxanneCor:
### Purpose of this PR
Draw Shape Tool
This tool lets the user draw a cube which will define the bounding box for any shape in it. You can change the type of the shape and its parameter whenever you like, even after completing it. You can also change the bounding box and its rotation afterwards.

### Comments to Reviewers
A part of the code is Karl's :)

How to use: Click on the New Shape menu, this will open the overlay to create a new shape. Click to create the last drawn shape or Click and Hold to draw a cube containing the shape. You can change the parameters of the shape in the overlay or in  the inspector.
You also have access to the Edit Shape Tool once you have created your shape. In this tool, you can change the size of the bounding box by dragging the dots on the faces or change the orientation by dragging the edges.

(This is for now a review for the code, I will add Joel and Gabriel once I've finished correcting some minor bugs)

### Bugfixes
Adding bugfix for :
Fogbugz : **https://fogbugz.unity3d.com/f/cases/1296520/**


Runtime/Shapes/Shape.cs
```diff
@@ -0,0 +1,13 @@
+#if UNITY_EDITOR
+using System.Reflection;
+using UnityEditor;
+#endif
```
neilunity, line 4: these using don't look necessary

Tests/Runtime/MeshOps/DeleteElementsTests.cs
```diff
@@ -6,18 +6,29 @@
...
+using System.Collections.Generic;
 
 static class DeleteElementsTests
 {
     static System.Random s_Random = new System.Random();
 
-    static ShapeType[] shapeTypes
-    {
-        get { return (ShapeType[])typeof(ShapeType).GetEnumValues(); }
+    public static List<Type> shapeTypes {
```
neilunity, line 15: this code is repeated 3 times below, could you put it in a function and use typecache ?

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -0,0 +1,375 @@
...
+        {
+            InitNewShape();
+            EditorTools.EditorTools.activeToolChanged += ActiveToolChanged;
+            m_ShapeTitle = new GUIContent("Draw Shape");
+            m_ShapeTypesPopupContent = m_AvailableShapeTypes.Select(x => x.Name).ToArray();
+        }
+
+        void InitNewShape()
+        {
+            m_Shape = new GameObject().AddComponent<ShapeComponent>();
```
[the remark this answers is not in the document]
RoxanneCor: The alternative would be this `new GameObject("Shape", typeof(ShapeComponent)).GetComponent<ShapeComponent>();`
I personnaly think it's cleaner without the GetComponent

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -0,0 +1,375 @@
...
+            m_Shape.gameObject.hideFlags = HideFlags.HideAndDontSave;
+            m_Shape.hideFlags = HideFlags.None;
+            m_Shape.SetShape(CreateShape(activeShapeType));
+            m_Object = new SerializedObject(m_Shape);
+        }
+
+        void OnDisable()
+        {
+            if (m_Shape.gameObject.hideFlags == HideFlags.HideAndDontSave)
```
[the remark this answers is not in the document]
RoxanneCor: Yes, when a shape is being drew. After the first click it's `HideFlags.None` and it goes back to `HideFlags.HideAndDontSave `only after the final click (when the shape is completed and a new hidden one is created).

Editor/EditorCore/EditShapeTool.cs
```diff
@@ -0,0 +1,374 @@
...
+            var faces = new FaceData[6];
+            var edges = new EdgeData[4];
+
+            // +X
+            var pos = m_BoundsHandle.center - new Vector3(extents.x, 0, 0);
+            edges[0] = new EdgeData(new Vector3(-extents.x, extents.y, extents.z), new Vector3(-extents.x, -extents.y, extents.z));
+            edges[1] = new EdgeData(new Vector3(-extents.x, extents.y, extents.z), new Vector3(-extents.x, extents.y, -extents.z));
+            edges[2] = new EdgeData(new Vector3(-extents.x, -extents.y, -extents.z), new Vector3(-extents.x, -extents.y, extents.z));
+            edges[3] = new EdgeData(new Vector3(-extents.x, -extents.y, -extents.z), new Vector3(-extents.x, extents.y, -extents.z));
+            faces[0] = new FaceData(pos, xAxis, (EdgeData[])edges.Clone());
```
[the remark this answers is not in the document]
RoxanneCor: Because even if the data was a struct the list was passed by reference. But I changed all that part anyway

Tests/Runtime/MeshOps/DeleteElementsTests.cs
RoxanneCor: I can't use TypeCache since it would need UnityEditor and we don't want that in those cases, but I'll put it in a global function

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -0,0 +1,375 @@
...
+            {
+                DestroyImmediate(m_Shape.gameObject);
+                InitNewShape();
+            }
+            EditorTools.EditorTools.activeToolChanged -= ActiveToolChanged;
```
lopezt-unity, line 80: If I'm not mistaken, EditorTools.EditorTools will not be valid anymore in Unity 2020.2, you may use something as follows to handle the different unity versions :

```
#if UNITY_2020_2_OR_NEWER
using ToolManager = UnityEditor.EditorTools.ToolManager;
#else
using ToolManager = UnityEditor.EditorTools.EditorTools;
#endif

```

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -0,0 +1,375 @@
...
+
+            if (m_Bounds.size.sqrMagnitude < .01f)
+                return;
+
+            // The sphere doesn't like being built before its height is set
+            if (m_Shape.shape is Sphere && System.Math.Abs(m_Bounds.size.y) < 0.01f)
+                return;
+
+            bool init = false;
+            if (!m_Shape.isInit)
```
lopezt-unity, line 137: Is there a special reason to use both init and m_Shape.isInit ? Seems pretty much the same to me here :)

RoxanneCor: I only want to enter the `if(init)` once, when it has just been init. I'll rename to make it more clear :)

Editor/ShapeParameters.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace UnityEditor.ProBuilder
+{
+    internal static class ShapeParameters
+    {
+        static Dictionary<string, Shape> s_Prefs = new Dictionary<string, Shape>();
+
+        static ShapeParameters()
+        {
+            var types = typeof(Shape).Assembly.GetTypes();
```
antoine-unity, line 15: The TypeCache should be used instead: https://docs.unity3d.com/ScriptReference/TypeCache.GetTypesDerivedFrom.html

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -0,0 +1,375 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Calculates the rotation angles to give a shape depending on the orientation we started drawing it
+        /// </summary>
+        /// <param name="diff">Difference between point A and point B</param>
+        /// <returns></returns>
+        Quaternion ToRotationAngles(Vector3 diff)
```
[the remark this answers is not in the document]
karl-: Why is this necessary in the first place? It looks to me like it's just "nicifying" the euler values.

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -0,0 +1,380 @@
...
+                            CancelShape();
+                        else
+                            AdvanceInputState();
+                        break;
+
+                    }
+            }
+        }
+
+        public static ProBuilderMesh CreateLastShape(Vector3 defaultSize)
```
karl-: I still find this a bit odd, where does `defaultSize` get used really?

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -0,0 +1,380 @@
...
+            if (s_Size == Vector3.zero)
+                s_Size = defaultSize;
+            var type = activeShapeType;
+            var shape = ShapeGenerator.CreateShape(type).GetComponent<ShapeComponent>();
+            ShapeParameters.SetToLastParams(ref shape.shape);
+            UndoUtility.RegisterCreatedObjectUndo(shape.gameObject, "Create Shape");
+
+            Bounds bounds = new Bounds(Vector3.zero, s_Size);
+            shape.Rebuild(bounds, Quaternion.identity);
+            shape.mesh.SetPivot(PivotLocation.Center);
```
karl-, line 301: This should be handled in `EditorUtility.InitObject`. The `PivotLocation` is a user preference, and not always center.

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -0,0 +1,375 @@
...
+            m_Bounds.center = ((m_OppositeCorner + m_Origin) * .5f) + m_Plane.normal * (height * .5f);
+            m_Bounds.size = new Vector3(fo, height, ri);
+            m_Rotation = Quaternion.identity;
+        }
+
+        void DrawBoundingBox()
+        {
+            using (new Handles.DrawingScope(new Color(.2f, .4f, .8f, 1f)))
+            {
+                EditorHandleUtility.PushMatrix();
```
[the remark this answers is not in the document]
karl-: yes, the `EditorHandleUtility` is a holdover from legacy probuilder code as antoine noted.

Editor/EditorCore/EditorHandleUtility.cs
```diff
@@ -511,24 +511,35 @@ internal static void PopMatrix()
...
-            if (GetPlaneFromPickedObject(mousePosition, out plane))
-                return plane;
+            var nrm = plane.normal;
 
-            if(GetPlaneFromProGridsAxis(mousePosition, out plane))
-                return plane;
+            bitangent = Vector3.Cross(nrm, Vector3.Cross(nrm, nrm == Vector3.right ? Vector3.forward : Vector3.right));
 
-            return GetPlaneFromCameraDirection();
+            return new SimpleTuple<Plane, Vector3>(plane, bitangent);
```
karl-, line 539: Looks like `GetPlaneFromCameraDirection` is omitted from the process here. Was that intentional?

Editor/EditorCore/EditorHandleUtility.cs
```diff
@@ -554,6 +565,9 @@ static bool GetPlaneFromPickedObject(Vector2 mousePosition, out Plane plane)
                         plane = new Plane(
                             go.transform.TransformDirection(hit.normal),
                             go.transform.TransformPoint(hit.point));
+                        var forward = go.transform.forward;
+                        var dot = Vector3.Dot(plane.normal, forward);
+                        bitangent = Mathf.Abs(dot) > .9f ? go.transform.up : forward;
```
karl-, line 570: this can result in a bitangent that is not parallel to the surface of the hit face. you'll need to do something like 

```
bitangent = nrm * (nrm *( |dot| > .9f ? transform.up : transform.forward ))
```

(where * is cross product, |x| is abs)

Editor/EditorCore/EditorHandleUtility.cs
```diff
@@ -562,13 +576,16 @@ static bool GetPlaneFromPickedObject(Vector2 mousePosition, out Plane plane)
                     UnityEngine.RaycastHit hit;
                     if (Physics.Raycast(HandleUtility.GUIPointToWorldRay(mousePosition), out hit))
                     {
-                        plane = new Plane(hit.normal,hit.point);
+                        plane = new Plane(hit.normal, hit.point);
+                        var forward = go.transform.forward;
+                        bitangent = Mathf.Abs(Vector3.Dot(plane.normal, forward)) > .9f ? go.transform.up : forward;
```
karl-, line 581: same note, this can result in a non-parallel bitangent

Editor/EditorCore/EditorUtility.cs
```diff
@@ -292,16 +292,18 @@ internal static void MoveToActiveScene(GameObject gameObject)
         /// Initialize this object with the various editor-only parameters, and invoke the object creation callback.
         /// </summary>
         /// <param name="pb"></param>
-        internal static void InitObject(ProBuilderMesh pb)
+        internal static void InitObject(ProBuilderMesh pb, bool snapToGrid = true)
```
karl-, line 295: `InitObject` is intended to apply the various user preferences, including snapping. Why are we passing `snapToGrid` instead of just checking if snapping is enabled?

Editor/EditorCore/EditorUtility.cs
```diff
@@ -292,16 +292,18 @@ internal static void MoveToActiveScene(GameObject gameObject)
...
         {
             MoveToActiveScene(pb.gameObject);
             GameObjectUtility.EnsureUniqueNameForSibling(pb.gameObject);
             ScreenCenter(pb.gameObject);
-            SetPivotLocationAndSnap(pb);
+            
+            if(snapToGrid)
```
karl-, line 301: The name `TrySnapToGrid` implies (to my ear) that it is already doing the snap enabled check. Why is `snapToGrid` also tested?

Editor/EditorCore/EditorUtility.cs
```diff
@@ -337,7 +339,8 @@ internal static void InitObject(ProBuilderMesh pb)
                 meshCreated(pb);
         }
 
-        internal static void SetPivotLocationAndSnap(ProBuilderMesh mesh)
+        // Snap new meshes to the grid if the preference is enabled, or progrids is active
+        internal static void TrySnapToGrid(ProBuilderMesh mesh)
```
karl-, line 343: 👍 to renaming this now that pivot isn't set here

Editor/EditorCore/ShapeComponentEditor.cs
```diff
@@ -0,0 +1,106 @@
...
+using UnityEngine.ProBuilder;
+using UnityEngine.UIElements;
+
+namespace UnityEditor.ProBuilder
+{
+    [CustomEditor(typeof(ShapeComponent))]
+    public class ShapeComponentEditor : Editor
+    {
+        private ShapeComponent m_ShapeComponent;
+        private IMGUIContainer m_ShapeField;
```
karl-, line 15: style guide - omit redundant access modifiers

Editor/EditorCore/ShapeComponentEditor.cs
```diff
@@ -0,0 +1,106 @@
...
+        SerializedProperty m_shape;
+        static string[] s_ShapeTypes;
+        static TypeCache.TypeCollection s_AvailableShapeTypes;
+
+        static Pref<int> s_ActiveShapeIndex = new Pref<int>("ShapeBuilder.ActiveShapeIndex", 0);
+
+        static ShapeComponentEditor()
+        {
+            s_AvailableShapeTypes = TypeCache.GetTypesDerivedFrom<Shape>();
+            s_ShapeTypes = s_AvailableShapeTypes.Select(x => x.ToString()).ToArray();
```
karl-, line 26: Assuming this is intended to be a menu item, should probably pass these through `ObjectNames.NicifyVariableName`

Editor/EditorCore/ShapeComponentEditor.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        private void OnEnable()
+        {
+            m_ShapeComponent = target as ShapeComponent;
+            m_shape = serializedObject.FindProperty("shape");
+            var fullName = m_shape.managedReferenceFullTypename;
+            var typeName = fullName.Substring(fullName.LastIndexOf(' ') + 1);
+
+            Type type = null;
+            foreach (var shapeType in s_AvailableShapeTypes)
```
karl-, line 37: It would be preferable to avoid string comparison here

Editor/EditorCore/ShapeComponentEditor.cs
```diff
@@ -0,0 +1,106 @@
...
+        public override void OnInspectorGUI()
+        {
+            DrawShapeGUI((ShapeComponent)target, serializedObject);
+        }
+
+        static Shape CreateShape(Type type)
+        {
+            var shape = Activator.CreateInstance(type) as Shape;
+            ShapeParameters.SetToLastParams(ref shape);
+            return shape;
```
karl-, line 61: I think I saw this function somewhere already?

Editor/MenuActions/Editors/OpenShapeEditor.cs
```diff
@@ -22,25 +24,24 @@ sealed class OpenShapeEditor : MenuAction
...
-        protected override MenuActionState optionsMenuState
-        {
+        protected override MenuActionState optionsMenuState {
             get { return MenuActionState.VisibleAndEnabled; }
         }
 
         public override ActionResult DoAction()
         {
-            ShapeEditor.CreateActiveShape();
+            DrawShapeTool.CreateLastShape(Vector3.one * 100);
```
karl-: Why is this `Vector3.one * 100` ?

Editor/MenuActions/Editors/OpenShapeEditor.cs
```diff
@@ -22,25 +24,24 @@ sealed class OpenShapeEditor : MenuAction
...
             return new ActionResult(ActionResult.Status.Success, "Create Shape");
         }
 
         protected override void DoAlternateAction()
         {
-            ShapeEditor.MenuOpenShapeCreator();
+            ProBuilderEditor.selectMode = SelectMode.Object;
+            ToolManager.SetActiveTool<DrawShapeTool>();
```
karl-, line 44: nice 👍

Runtime/Core/ProBuilderMeshFunction.cs
```diff
@@ -211,7 +211,7 @@ public static ProBuilderMesh Create(IEnumerable<Vector3> positions, IEnumerable<
             return mesh;
         }
 
-        void GeometryWithPoints(Vector3[] points)
+        internal void GeometryWithPoints(Vector3[] points)
```
karl-, line 214: what new code is using this that required it to be internal? my preference would be to deprecate `GeometryWithPoints` in favor of any of the other initialization functions, but if this is just moving code that's already using then it's fine

Runtime/Core/ProBuilderMeshFunction.cs
```diff
@@ -240,6 +240,19 @@ void GeometryWithPoints(Vector3[] points)
             Refresh();
         }
 
+        /// <summary>
+        /// Rebuilds the mesh with different vertices positions but same faces
+        /// </summary>
+        /// <param name="points">New vertices positions</param>
+        internal void ReplaceVertices(Vector3[] points)
```
karl-, line 247: couldn't this just use the `positions` setter? i'm assuming that this is used just to handle resizing the bounds, in which case there isn't any need to rebuild the shared vertex cache or lookup.

likewise, no need for a complete mesh rebuild, you can just upload the positions directly to the mesh and avoid a lot of unnecessary overhead.

Runtime/Shapes/ShapeComponent.cs
```diff
@@ -0,0 +1,144 @@
+﻿using System;
+
+namespace UnityEngine.ProBuilder
+{
+    [RequireComponent(typeof(ProBuilderMesh))]
+    public class ShapeComponent : MonoBehaviour
+    {
+        [SerializeReference]
+        public Shape shape = new Cube();
```
karl-, line 9: there should not be any public fields in classes - they need to be exposed through properties

Runtime/Shapes/ShapeComponent.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        public bool isInit;
```
karl-, line 11: noted above, let's see if we can figure out a way to avoid needing an init flag in this class

Runtime/Shapes/ShapeComponent.cs
```diff
@@ -0,0 +1,144 @@
...
+            get { return m_Size; }
+            set { m_Size = value; }
+        } 
+
+        public ProBuilderMesh mesh {
+            get { return m_Mesh == null ? m_Mesh = GetComponent<ProBuilderMesh>() : m_Mesh; }
+        }
+
+        // Bounds where center is in world space, size is mesh.bounds.size
+        internal Bounds meshFilterBounds {
```
karl-, line 44: I have vague memories of this, but can you remind me why we need a bounds that has the center in world space with size in model space?

Runtime/Shapes/ShapeComponent.cs
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        public void SetShape(Shape shape)
+        {
+            this.shape = shape;
+            Rebuild();
+        }
+
+        void FitToSize()
+        {
```
karl-, line 73: this would be a useful function to have in a public utility class

Runtime/Shapes/ShapeComponent.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            mesh.ToMesh();
+            mesh.Rebuild();
+        }
+
+        /// <summary>
+        /// Set the rotation of the Shape to a given set of eular angles, then rotates it
+        /// </summary>
+        /// <param name="eulerAngles">The angles to rotate by</param>
+        public void SetRotation(Quaternion angles)
```
karl-, line 104: something to consider, it may be worth naming this something that reflects that it is rotating within the dimensions of the bounds

Runtime/Shapes/ShapeComponent.cs
```diff
@@ -0,0 +1,144 @@
...
+            rotationQuaternion = angles;
+            ApplyRotation(rotationQuaternion);
+            FitToSize();
+        }
+
+        /// <summary>
+        /// Rotates the Shape by a given set of eular angles
+        /// </summary>
+        /// <param name="eulerAngles">The angles to rotate by</param>
+        public void Rotate(Quaternion rotation)
```
karl-, line 115: XML docs note that this is rotating by a delta in Euler angles, but it accepts a quaternion

Runtime/Shapes/ShapeComponent.cs
```diff
@@ -0,0 +1,144 @@
...
+            shape.RebuildMesh(mesh, size);
+
+            var origVerts = mesh.positionsInternal;
+
+            for (int i = 0; i < origVerts.Length; ++i)
+            {
+                origVerts[i] = rotation * origVerts[i];
+            }
+            mesh.mesh.vertices = origVerts;
+            mesh.ReplaceVertices(origVerts);
```
karl-, line 141: you could omit the `mesh.ReplaceVertices` call here.

Runtime/Core/ShapeGenerator.cs
```diff
@@ -1,71 +1,8 @@
-using UnityEngine;
-using System.Collections.Generic;
 using UnityEngine.ProBuilder.MeshOperations;
+using System;
 
 namespace UnityEngine.ProBuilder
 {
-    /// <summary>
```
karl-, line 7: because this is public API, it needs to go through a deprecation release before we can remove it

Editor/EditorCore/DrawShapeTool.cs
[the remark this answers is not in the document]
RoxanneCor: Yep, removed the` InitShape()`

Editor/EditorCore/FreeFormBoxTool.cs
```diff
@@ -0,0 +1,225 @@
...
+        {
+            if(!ToolManager.IsActiveTool(this))
+                return;
+#endif
+            foreach(var obj in targets)
+            {
+                var pbmesh = obj as ProBuilderMesh;
+                pbmesh.SetPivot(pbmesh.transform.position + pbmesh.mesh.bounds.center);
+
+                m_Modifications.Add(pbmesh, new InternalModification(pbmesh.positionsInternal));
```
karl-: Why is the `m_Modifications` dictionary necessary?

Editor/EditorCore/FreeFormBoxTool.cs
```diff
@@ -0,0 +1,225 @@
...
+
+        void OnObjectSelectionChanged()
+        {
+            if(!ToolManager.IsActiveTool(this))
+                return;
+
+            foreach(var mesh in m_Modifications.Keys)
+            {
+                if(!targets.Contains((Object) mesh))
+                    EditorApplication.delayCall += () => m_Modifications.Remove(mesh);
```
karl-, line 104: Why is this in a `delayCall`?

Editor/EditorCore/FreeFormBoxTool.cs
```diff
@@ -0,0 +1,225 @@
...
+            Bounds bounds = mesh.mesh.bounds;
+
+            var origVerts = new Vector3[mesh.vertexCount] ;
+            Array.Copy(currentModification.vertices, origVerts, mesh.vertexCount);
+
+            for (int i = 0; i < origVerts.Length; ++i)
+                origVerts[i] = currentModification.rotation * origVerts[i] + bounds.center;
+
+            mesh.mesh.vertices = origVerts;
+            mesh.ReplaceVertices(origVerts);
```
karl-, line 195: why are we copying the vertex array, then re-assigning it? this should just access `mesh.positionsInternal`

Editor/EditorCore/ShapeComponentEditor.cs
```diff
@@ -0,0 +1,131 @@
...
+        Type[] m_AvailableShapeTypes;
+
+        int m_ActiveShapeIndex = 0;
+
+        const string k_dialogTitle = "Shape reset";
+        const string k_dialogText = "The current shape has been edited, you will loose all modifications.";
+
+        ShapeComponentEditor()
+        {
+            m_AvailableShapeTypes =  TypeCache.GetTypesWithAttribute<ShapeAttribute>().Where(t => t.BaseType == typeof(Shape)).ToArray();
```
karl-, line 32: Looks like this would break if there is an abstract implementation of shape, ex:

```
abstract class ProceduralShape : Shape {}
class ProceduralCube : ProceduralShape {}
```

Editor/EditorCore/VertexManipulationTool.cs
```diff
@@ -267,6 +268,10 @@ protected void BeginEdit(string undoMessage)
 
             foreach (var mesh in selection)
             {
+                ShapeComponent shape;
+                if(mesh.gameObject.TryGetComponent(out shape))
+                    shape.edited = true;
```
karl-, line 273: This isn't a reliable way to set the edited flag. In fact I don't know if we have a good way to do this at the moment that can accurately handle all cases. It may require us to add some sort of hash or ID to the `ProBuilderMesh` class that can be used to track changes.

karl- (changes requested):

Editor/EditorShapeUtility.cs
```diff
@@ -0,0 +1,340 @@
...
+        //         ( boxBoundsHandle.center - activeBoundsState.boundsHandleValue.center ).normalized;
+        //     Vector3 sizeDiffSign = ( boxBoundsHandle.size - activeBoundsState.boundsHandleValue.size ).normalized;
+        //     Vector3 globalSign = Vector3.Scale(centerDiffSign, sizeDiffSign);
+        //     //Set the center to the right position
+        //     Vector3 center = activeBoundsState.boundsHandleValue.center +
+        //                      Vector3.Scale(( snappedHandleSize - activeBoundsState.boundsHandleValue.size ) / 2f,
+        //                          globalSign);
+        //     //Set new Bounding box value
+        //     activeBoundsState.boundsHandleValue = new Bounds(center, snappedHandleSize);
+        // }
```
karl-, line 229: remove unused code

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -0,0 +1,378 @@
...
+            if (s_ActiveShapeIndex < 0)
+                throw new Exception("type must inherit UnityEngine.ProBuilder.Shape");
+
+            if (m_Shape != null)
+            {
+                if (m_InputState != InputState.SelectPlane)
+                    RebuildShape();
+            }
+        }
+
```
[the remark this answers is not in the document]
karl-: I don't know, this wasn't something I wrote 🤷

Editor/EditorCore/EditorHandleUtility.cs
```diff
@@ -519,14 +519,15 @@ static Vector3 GetBitangent(Vector3 planeNormal)
         /// Get a plane suitable for mouse input in a scene view.
         /// </summary>
         /// <param name="mousePosition"></param>
+        /// <param name="alignOnGrid"></param>
         /// <returns></returns>
-        internal static Plane FindBestPlane(Vector2 mousePosition)
+        internal static Plane FindBestPlane(Vector2 mousePosition, bool alignOnGrid = false)
         {
-            var res = FindBestPlaneAndBitangent(mousePosition);
+            var res = FindBestPlaneAndBitangent(mousePosition, alignOnGrid);
```
lopezt-unity: @karl- 
I added a boolean to trace the Shape and Polyshape on the sceneview grid if visible rather than at a random height.
Is this boolean a necessity or should this be the default behavior ?
If I remember it was a remark from @gabrielw-us originally :)

karl-: yes this should just be the default behaviour, no align on grid argument necessary

Editor/EditorCore/VertexManipulationTool.cs
lopezt-unity: I Added this remark on the Shape tool's Jira

karl- (approved):

PR #331, opened 2020-08-13
Menutoggle

lopezt-unity:
### Purpose of this PR

Adding MenuToggle, inheriting from MenuAction.
MenuToggle allows to have actions that last time and have a start and an end, whereas MenuAction is more a simple trigger.

### Comments to Reviewers
This PR is necessary for future functionalities: Shape Tool, Cut tool, new PolyShape


lopezt-unity:
This was originally to share with Roxanne, to integrate in the Shape Tool.
However, it' going to be different now, I'll see with Karl if this PR is needed, but it might not anymore.

PR #333, opened 2020-08-20
Probuilder Editor refactoring and upgrade to Editor Tool

lopezt-unity:
### Purpose of this PR

Upgrade the current version of Probuilder 4.4.0 to 5.0.0 :
- VertexManipulationTool is now inheriting from EditorTool 

### Links

**Jira:** https://jira.unity3d.com/browse/STO-1476

### Comments to Reviewers


Editor/EditorCore/ProBuilderEditor.cs
```diff
@@ -528,47 +535,69 @@ void SetIsUtilityWindow(bool isUtilityWindow)
...
-            if (s_EditorTools.TryGetValue(typeof(T), out tool))
-                return tool;
-            tool = new T();
-            s_EditorTools.Add(typeof(T), tool);
-            return tool;
+            if(EditorToolManager.activeTool is VertexManipulationTool)
+                formerTool = (VertexManipulationTool)EditorToolManager.activeTool;
+            ToolManager.SetActiveTool(CreateInstance<T>());
+            if(formerTool != null)
+                DestroyImmediate(formerTool);
```
karl-, line 571: `ToolManager.SetActiveTool<T>();`

`ToolManager.RestorePreviousPersistentTool`

Editor/EditorCore/ProBuilderEditor.cs
```diff
@@ -684,19 +713,16 @@ void OnSceneGUI(SceneView sceneView)
                     break;
             }
 
+#if !UNITY_2020_2_OR_NEWER
             // Overrides the toolbar transform tools
-            if (Tools.current != Tool.None && Tools.current != m_CurrentTool)
+            if (Tools.current != Tool.None && Tools.current != Tool.Custom && Tools.current != m_CurrentTool)
```
karl-, line 718: Use `ToolManager`

Editor/EditorCore/VertexManipulationTool.cs
```diff
@@ -214,27 +213,25 @@ protected float GetSnapValueForAxis(Vector3Mask axes)
...
 
-        static void Init()
+        void OnEnable()
         {
-            if (s_Initialized)
-                return;
-            s_Initialized = true;
             s_VertexDragging = typeof(Tools).GetField("vertexDragging", BindingFlags.NonPublic | BindingFlags.Static);
             s_FindNearestVertex = typeof(HandleUtility).GetMethod("FindNearestVertex",
                     BindingFlags.Static | BindingFlags.NonPublic | BindingFlags.Instance);
```
karl-, line 220: Not related, but we should have access to these functions now

Editor/EditorCore/ProBuilderEditor.cs
karl-: Or remove?

Editor/EditorCore/ProBuilderEditor.cs
lopezt-unity: So using the Tool as a Singleton (with ToolManager.SetActiveTool<T>();) and not destroying it after use let the tool appear in the tool history.
image

However, tools from the history cannot be selected has the main tools are the ones that are considered.
So, what is the best behaviour to get on this ?

karl-: This seems like something we should fix on trunk in the `EditorToolManager`

lopezt-unity: Is it something to fix? 
As the tool is a singleton and not destroy, it's normal that it appears in there right ?

karl-: That list is curated, and should only be showing tools that have an `EditorTool` attribute. 

I think it would also make sense to cull tools that resolve to BuiltIn tools, but I could see arguments to the contrary there.

lopezt-unity: Agreed with the `EditorTool` attribute

Editor/EditorCore/VertexManipulationTool.cs
```diff
@@ -137,8 +138,8 @@ static void SyncPivotRotation()
         internal static Pref<bool> s_ExtrudeEdgesAsGroup = new Pref<bool>("editor.extrudeEdgesAsGroup", true);
         internal static Pref<ExtrudeMethod> s_ExtrudeMethod = new Pref<ExtrudeMethod>("editor.extrudeMethod", ExtrudeMethod.FaceNormal);
 
-        Vector3 m_HandlePosition;
-        Quaternion m_HandleRotation;
+        protected Vector3 m_HandlePosition;
```
antoine-unity, line 141: If I remember correctly, the coding style for protected value is thee same as public. @karl- can you confirm?

JoelFortin (approved):
Checked with Thomas on what was already tested. approving.

PR #339, opened 2020-09-03
Merge function to public + Updating MenuAction samples

lopezt-unity:
### Purpose of this PR

A user wanted to make is own action to merge faces from an edge selection.
We thus exposed MergeElements class.
We added the SoftDeleteEdgesAction needed by the user to the samples as well.

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1271905/


Runtime/MeshOperations/MergeElements.cs
```diff
@@ -8,7 +8,7 @@ namespace UnityEngine.ProBuilder.MeshOperations
     /// <summary>
     /// Merging faces together.
     /// </summary>
-    static class MergeElements
+    public sealed class MergeElements
```
karl-, line 11: This could use a more descriptive summary for the API documentation. We'll also want to have Amy (@unitALG ) review the existing documentation in this class now that it is becoming public.

Samples~/Editor/SoftDeleteEdgesAction.cs
```diff
@@ -0,0 +1,98 @@
...
+				if(pbMesh.selectedEdgeCount > 0)
+				{
+					var selectedEdges = pbMesh.selectedEdges;
+					
+					edgesBounds = selectedEdges.Select(edge => new SimpleTuple<Vector3,Vector3>(pbMesh.positions[edge.a], pbMesh.positions[edge.b])).ToList();
+					
+					for(int i= 0; i < edgesBounds.Count; i++)
+					{	
+						List<Face> facesToMerge = pbMesh.faces.Where(face => face.edges.ToList().Exists(e => IsEdgeEquivalent(pbMesh, e, edgesBounds[i]))).ToList();
+						Face f = MergeElements.Merge(pbMesh, facesToMerge);
```
karl-, line 71: We could make `ElementSelection.GetNeighborFaces` public and use that function instead. It would be a good idea to add a non-allocating overload of that function as well.

antoine-unity: It would be nice if we can dodge having to ToList the ReadOnlyArray by having an internal function that can use edgesInternal instead. Also, having a public and optimized function that gets face neighbors with a NonAlloc result overload seems like a good idea. 

In context, I think the line is perfectly fine, but considering that this is a common enough thing to do that people would probably copy/paste this from the sample and not understand why their stuff, in a more critical path, might be slow.  I don't think we should be optimizing the samples but any stuff that requires linq to be clean could point to a lack in our public API.

lopezt-unity: @karl- and @antoine-unity 
I did try to make this as you were saying, feel free to let me know your thoughts on this.
The code is way more dense than before as this 'simple' case a little more complicated than expected at first sight, OR I might be missing some info and in that case, please give me an update on this :) :
- we cannot merge every single couple of face designated by an edge directly as it'll mess up the rest of the selection => _selected edges will be modified by the merge_
- we don't want to merge all faces designated by the edges together => _you want to remove edges between faces, not merge all faces together_
- we do want to create a group of face to merge together if edge selection contains enough edges to connect them

This new method take all of this into account.
The previous one was doing the same too by reasoning on Vertices positions (Vector3) instead of indexes

Let me know your thoughts on that and if you have a simpler solution to handle this :)

antoine-unity (approved):

karl- (approved):

JoelFortin (approved):
LGTM. Peer reviewed this with Thomas.
Eventually, I'd like a little option window for that action where we could decide if we want to remove the faces/or keep them ... also remove floating vertices after transformation or not.

PR #340, opened 2020-09-10
Bug fix collection

karl-:
### Purpose of this PR
```
- [case: 1276085] Fixed `UV Actions` window allowing resize events outside of the containing window.
- [case: 1276074] Fixed a case where `Fit UVs` action could result in `NaN` values.
- Fixed rect selection not working with Universal Render Pipeline.
```
### Links

**Fogbugz:** 

- https://fogbugz.unity3d.com/f/cases/1276085/
- https://fogbugz.unity3d.com/f/cases/1276074/


Runtime/Core/SelectionPickerRendererStandard.cs
```diff
@@ -91,7 +80,16 @@ internal class SelectionPickerRendererStandard: ISelectionPickerRenderer
...
+
+                // URP does not support replacement shaders or custom passes from code, so for now it is necessary to
+                // force the pipeline to built-in when rendering. In editor it may be possible to use Handles.DrawCamera
+                // to avoid disposing and re-assigning the pipeline, as the RenderEditorCamera function has some logic
+                // that switches rendering path if replacement shaders are in use, but I wasn't able to get that
+                // approach to work without also requiring that the drawing happen during a repaint event.
+                var currentRenderPipeline = GraphicsSettings.renderPipelineAsset;
+                GraphicsSettings.renderPipelineAsset = null;
                 renderCam.RenderWithShader(shader, tag);
+                GraphicsSettings.renderPipelineAsset = currentRenderPipeline;
```
lopezt-unity, line 92: Would that also fix the problem we have in probuilder with vertex selection in URP ?

lopezt-unity (approved):

Runtime/Core/SelectionPickerRendererStandard.cs
karl-: Yes, it's a workaround until we can implement it properly

johnson512n28: This workaround does not appear to work around in 2020.1.2f1.

JoelFortin (approved):
Peer reviewed with Karl. LGTM

PR #341, opened 2020-09-17
Take into account the new default parent object feature

antoine-unity:
### Purpose of this PR

2021.1 now has a feature to set an object in each open scenes as the default parent when adding object.
The shape creator now takes this into account.

image

### Links

**Fogbugz:** [case#1277788](https://fogbugz.unity3d.com/f/cases/1277788/)


Editor/EditorCore/EditorUtility.cs
```diff
@@ -250,11 +250,20 @@ internal static bool HasStaticFlag(this GameObject go, StaticEditorFlags flags)
         }
 
         /// <summary>
-        /// Move a GameObject to the active scene, where active scene may be a prefab stage.
+        /// Move a GameObject to the proper active root.
+        /// Checks if a default parent exists, otherwise it adds the object as a root of the active scene, which can be a prefab stage.
         /// </summary>
         /// <param name="gameObject"></param>
-        internal static void MoveToActiveScene(GameObject gameObject)
+        internal static void MoveToActiveRoot(GameObject gameObject)
```
lopezt-unity, line 257: This method seems to be used in other classes (at least in `ShapeEditor`) but I don't see modifications in other places after this renaming.

Editor/EditorCore/EditorUtility.cs
```diff
@@ -250,11 +250,20 @@ internal static bool HasStaticFlag(this GameObject go, StaticEditorFlags flags)
...
         {
+#if UNITY_2021_1_OR_NEWER
```
lopezt-unity, line 259: Problem was raised in 2020.2.0b3, is this `SceneView.GetDefaultParentObjectIfSet` only available from 2021?

Editor/EditorCore/EditorUtility.cs
karl-: This could be made public as well

karl- (approved):

Editor/EditorCore/EditorUtility.cs
antoine-unity: @lopezt-unity good catch, was missing a file

antoine-unity: @karl- Might be better to make InitObject public

lopezt-unity (approved):

JoelFortin (approved):
messed around with the default parent option and LGTM.

PR #347, opened 2020-10-27
Fixed warning when modifying a PBMesh with particule effect using PBMesh as shape emitter

lopezt-unity:
### Purpose of this PR

Fixed warning when modifying a PBMesh with particule effect using PBMesh as shape emitter.
Mesh.Optimize was causing the mesh the be cleaned even when no vertices has been changed, causing the warning message.

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1281658/
**Jira:**

### Comments to Reviewers


JoelFortin (changes requested):
If the emitter's topology is modified while emitting particules ... we'll still get the warning in the console.

karl- (approved):

lopezt-unity:
> If the emitter's topology is modified while emitting particules ... we'll still get the warning in the console.

This is now handled, no warning message should be triggered by the particle system

JoelFortin (approved):
LGTM

PR #356, opened 2020-12-01
Tool Manager fix and improvement

meckzqz:
### Purpose of this PR

Removed presumably obsolete code resulting in the selection mode to be set to face selection both when launching the editor, and when entering or exiting play mode. In addition, the behaviour in ProBuilderToolManager's Dispose method where the selection mode is reset to object selection has been removed, allowing your selection mode to persist after entering and exiting play mode.


Editor/EditorCore/ProBuilderEditor.cs
```diff
@@ -49,7 +49,6 @@ public sealed class ProBuilderEditor : EditorWindow, IHasCustomMenu
         static ProBuilderToolManager toolManager => s_Instance != null ? s_Instance.m_ToolManager : null;
         internal EditorToolbar toolbar => m_Toolbar; // used by unit tests
         static ProBuilderEditor s_Instance;
-        static Pref<SelectMode> s_LastActiveSelectMode = new Pref<SelectMode>("editor.lastActiveSelectMode", SelectMode.Face);
```
karl-, line 52: Should add `"editor.lastActiveSelectMode"` to `PreferenceUpdater.DeleteObsoletePreferences` to clean up unused pref keys.

Editor/EditorCore/ProBuilderToolManager.cs
```diff
@@ -95,8 +95,6 @@ public void Dispose()
             m_IsDisposed = true;
             GC.SuppressFinalize(this);
 
-            SetSelectMode(SelectMode.Object);
```
lopezt-unity, line 98: Not sure this should be removed.
It ensures that Select mode (and in more recent version, the EditorTools Context) are restored when closing the ProBuilder editor window, insuring to restore GameObject context for the tools.
It sure set the selectmode to object when entering playmode but this is then updating by the probuilder editor.
What do you think @karl-  ?

lopezt-unity (changes requested):

Editor/EditorCore/ProBuilderEditor.cs
meckzqz: this change has now been made and pushed

Editor/EditorCore/ProBuilderToolManager.cs
karl-: Yes I think this needs to remain here.

JoelFortin (approved):

karl-:
Closing this for now, once PR feedback is addressed please feel free to re-open.

PR #345, opened 2020-10-16
Stability improvements

karl-:
### Purpose of this PR

Various stability improvements, mostly focused on `ProBuilderEditor` and tool integration.

- Fixed possible leak of `EditorToolbar` object.
- Fixed case where `EditorToolbar` could lose the `ProBuilderEditor` instance.
- Fixed issue where maximizing the `ProBuilder Editor` window and minimizing would break tool input.
- Fixed various cases where switching tools or selection modes could cause the `EditorTool` and `ProBuilderEditor` to lose connection.
- Re-enable UV upgrade tests with fix for Yamato failures
- Remove dead code
- Add unit tests for `EditorTool` and `EditorToolContext` integration

### Links

[ProBuilder handles do not update with tool changes](https://fogbugz.unity3d.com/f/cases/1285118/)
[Unable to select "Selection Tool" in the Scene View when ProBuilder window is maximized and minimized](https://fogbugz.unity3d.com/f/cases/1285054/)

PR #357, opened 2020-12-02
Fixing memory leak created by the constant refresh of the selected pbMesh

lopezt-unity:
### Purpose of this PR

Fixing a memory leak of ~4Mb per second using PB 5.0 with Unity >= 2021.1.

The problem was introduced since this commit :
https://github.com/Unity-Technologies/com.unity.probuilder/pull/345/commits/0fb88c56551d36f1e26d19989ce54509e4460239

The modification from 
```
                var mode = selectMode;
                mode = UI.EditorGUIUtility.DoElementModeToolbar(m_ElementModeToolbarRect, mode);
                if (mode != selectMode)
                    selectMode = mode;
```
to
```
                selectMode = UI.EditorGUIUtility.DoElementModeToolbar(m_ElementModeToolbarRect, selectMode);
```

caused constant calls to `Refresh()` on the selection and thus I tracked the leak to the call of the method `RebuildMeshHandle`. 
It seems that with newer version of Unity methods `MeshHandles.CreateFaceMesh` and `MeshHandles.CreateEdgeMesh` are leaking as it was not causing problem with previous Unity version.

To prevent this this fix reduces calls to `Refresh `to be done only when necessary and not every frame.

However, I guess we should have a detailed look in the Clear method as it seems it is not deallocating memory the way it should with lastest Unity versions.
`        [NativeMethod("Clear")]                 extern private void ClearImpl(bool keepVertexLayout);`

### Comments to Reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Bug Fixes
 
+- Fix for constant PB Mesh refresh causing Memory leak 
 - [case: 1254339] Correct offset when rendering UVs and correct export when UV Editor is docked.
 - Fixed PolyShape in Prefab mode: PolyShapeMode was not serialized when exiting prefab mode. Update point insertion visualization.
 - [case: 1259845] Fixed dimension overlay being hidden on playmode or reboot of the editor.
--- a/Editor/EditorCore/ProBuilderEditor.cs
+++ b/Editor/EditorCore/ProBuilderEditor.cs
@@ -190,8 +190,8 @@ namespace UnityEditor.ProBuilder
 
             set
             {
-                toolManager?.SetSelectMode(value);
-                Refresh();
+                if(toolManager != null && toolManager.SetSelectMode(value))
+                    Refresh();
             }
         }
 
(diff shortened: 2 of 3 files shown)


lopezt-unity:
@karl- I don't know how PB dlls were generated for these native calls. I could use your help on this :)

karl-:
the constant repaint was fixed on the shape drawing branch in a slightly different way https://github.com/Unity-Technologies/com.unity.probuilder/commit/a11993aa5b7fb77a58665c4fb274ffa4c11431ff

this doesn't fix the leak, it's only lessening the impact right?

antoine-unity (approved):

lopezt-unity:
@karl- yeah it's only reducing leak's impact.
I guess the real fix would be on trunk side.

I didn't check how it was handled on the shape branch, I don't know which one is better between the 2 fixes?

karl-:
my preference would be https://github.com/Unity-Technologies/com.unity.probuilder/commit/a11993aa5b7fb77a58665c4fb274ffa4c11431ff so that the editor will always refresh when the select mode changes, regardless of whether it's modified from `ProBuilderEditor` or `ProBuilderToolManager`

JoelFortin (approved):
LGTM

PR #361, opened 2020-12-16
Editor/filter set tooltip window position

karl-:
### Purpose of this PR
Fix tooltip window setting position every frame regardless of whether it has changed.

### Links

**Fogbugz:** 1266769
**Jira:**


lopezt-unity (approved):
LGTM

Just wondering, where was the actual problem with tooltip on Linux coming from? 
Was it only due to calling  `s_WindowRect = new Rect(0, 0, size.x, size.y);` too often?

karl-:
Yes, I guess this call was problematic with the way window events are handled on Linux.

PR #366, opened 2021-02-02
Added Missing Dependency Builtin Modules

antoine-unity:
### Purpose of this PR

Removing the module Particle System or IMGUI was causing compilation errors. Added both of those as dependencies. I validated that removing everything in the manifest file didn't cause any other errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Bug Fixes
+
+- Added Particle System and IMGUI modules as a dependency.
+
 ## [5.0.0-pre.10] - 2021-01-22
 
 ### Internal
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Bug Fixes
+
+- Added Particle System and IMGUI modules as a dependency.
```
karl-, line 12: @lopezt-unity can we make particle system an optional dependency?

AverageComet250: Not really as without it you can't compile.

gytis-zasciurinskas (approved):
Checked the package compilation with 2019.4.20f1, 2020.2.3f1, 2021.1.0b5 and 2021.2.0a3 and it has compiled properly in each stream. I have also done some quick modeling with 21.2.0a3 specifically, to see if ProBuilder was functioning as expected. Haven't noticed any issues, so PR looks good!

CHANGELOG.md
lopezt-unity: I added this as an optional dependency, it should be lighter.
Thanks Antoine for the help on this one.

lopezt-unity (approved):

karl- (approved):

PR #368, opened 2021-02-09
Bugfix/1312537 script stripping disable objects

lopezt-unity:
### Purpose of this PR

Fixing script stripping when the scene has disabled objects and the application is built.
Also fixing script stripping that was not taking the new ShapeComponents into account and creating errors in PB5.0.

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1312537/

### Comments to Reviewers


lopezt-unity:
@karl- , the problem was notified for PB4.4, should we backport the solution on the PB4.5 branch?

gytis-zasciurinskas (approved):
We did a peer review with Thomas for this. We have set up a small project to check if the scripts are stripped in the Build and everything is now working as expected. I will attach this project to the FogBugz ticket for future reference (as currently there was no repro project attached)
Also checked the use case where the Shape Renderer would be removed when Shape's scripts were stripped in the Editor (Using Tools -> ProBuilder ->  Action). This is now also working as expected.

Editor/EditorCore/StripProBuilderScripts.cs
```diff
@@ -125,5 +118,16 @@ public static void DoStrip(ProBuilderMesh pb)
...
+
+        static void DestroyProBuilderMeshAndDependencies(GameObject go, ProBuilderMesh pb)
+        {
+            DestroyImmediate(pb);
+
+            if (go.TryGetComponent(out Entity entity))
+                DestroyImmediate(entity);
+
+            if (go.TryGetComponent(out ShapeComponent shapeComponent))
+                DestroyImmediate(shapeComponent);
```
karl-, line 130: is this missing spline and polyshape components?

karl-: also, can this use the same function as what we have in `UnityScenePostProcessor`?

Tests/Editor/Actions/StripProBuilderScriptsTest.cs
```diff
@@ -45,4 +47,78 @@ public void StripProBuilderScripts_RemovesMeshAndBezierComponents()
...
+        Assume.That(go.GetComponent<ProBuilderMesh>() != null);
+        Assume.That(go.GetComponent<ShapeComponent>() != null);
+
+        StripProBuilderScripts.DoStrip(go.GetComponent<ProBuilderMesh>());
+
+        Assert.That(go.GetComponent<ProBuilderMesh>() == null);
+        Assert.That(go.GetComponent<ShapeComponent>() == null);
+
+        UObject.DestroyImmediate(go);
+    }
```
karl-, line 67: awesome to see tests added for this 🚀

Editor/EditorCore/StripProBuilderScripts.cs
lopezt-unity: I tried to put the scripts cleaning in a single method. 
However, each place using script stripping is doing this in a different way, which complicates this...

I also updated script stripping when exporting prefab and meshes.
They all call now the same method to ease future stripping extensions is needed.

Editor/MenuActions/Export/ExportAsset.cs
```diff
@@ -98,11 +99,7 @@ public static string ExportWithFileDialog(IList<ProBuilderMesh> meshes, ExportAs
                         res = ExportMesh(assetPath, pb, options.replaceOriginal, true);
 
                         if (options.replaceOriginal)
-                        {
-                            pb.preserveMeshAssetOnDestroy = true;
-                            Undo.DestroyObjectImmediate(pb);
-                        }
-
+                            StripProBuilderScripts.DestroyProBuilderMeshAndDependencies(pb.gameObject, pb, true);
```
lopezt-unity, line 102: @gytis-zasciurinskas All the modifications in this files come from a remark from @karl- .
Using Export>Export as mesh or prefab as causing error while performed on a gameobject with a ProBuilder Shape

karl- (approved):

gytis-zasciurinskas (approved):
We did a peer review with Thomas, the Export functionality now works correctly, no errors are now thrown
I1cxLubDPt

PR #369, opened 2021-02-09
Fix for material reverting when subdividing an edge

lopezt-unity:
### Purpose of this PR

Fixing problem occurring when subdividing an edge on a PB Mesh with multiple materials assigned to the faces.
The fix ensure the submesh index is kept when duplicating the face.

Before the fix:
materialProblem-before

After the fix:
materialProblem-after

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1311258/

### Comments to Reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Bug Fixes
 
+- [case: 1311258] Fixing material reverting when subdividing edge.
 - Added Particle System and IMGUI modules as a dependency.
 
 ## [5.0.0-pre.10] - 2021-01-22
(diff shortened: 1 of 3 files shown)


karl- (approved):

gytis-zasciurinskas (commented):
Tested with Legacy and HDRP for compatibility comparison. Subdivision now works as expected and materials no longer revert when an edge or a face is subdivided.

However, I was able to find an edge use case. The problem can still reproduce when the Cut tool is used.
igeFdDXgcr

lopezt-unity:
> Tested with Legacy and HDRP for compatibility comparison. Subdivision now works as expected and materials no longer revert when an edge or a face is subdivided.
> 
> However, I was able to find an edge use case. The problem can still reproduce when the Cut tool is used.

Just solved this to get a consistent behavior.

gytis-zasciurinskas (approved):
Rechecked the PR and the materials no longer revert when using the cut tool!:) Also, tested the behavior with Bevel, Extrude Faces, Triangulate Faces and Detach Faces tools and these worked fine too

PR #372, opened 2021-02-24
Bugfix/edges picking not working unity20.2

lopezt-unity:
### Purpose of this PR

Fixing edge selection using rect box selection that was broken with some Unity version.
The fix is using the same principle that the highlight shader is using : the edge picker shader is bringing edges a little closer to the camera plane to avoid z-fighting problem. This z-fighting was causing the problematic rendering in the texture.

However the problem was really inconsistent and reproduction was not always possible.

It seems the problem was coming from a problem with depth precision and was present on some Unity 20.2 versions.
Here is an example of the broken selection :
edge-selection-broken

Here is the result after applying the fix :
edge-selection-fixed

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1317148/

### Comments to Reviewers
This fix is only needed for a few Unity versions, but it's working for all of Unity version, so I would advice to use this new shader version for all Unity versions from now.

### DIFF
--- a/Content/Shader/EdgePicker.shader
+++ b/Content/Shader/EdgePicker.shader
@@ -43,7 +43,13 @@ CGPROGRAM
             v2f vert (appdata v)
             {
                 v2f o;
-                o.pos = UnityObjectToClipPos(v.vertex);
+                o.pos = float4(UnityObjectToViewPos(v.vertex.xyz), 1);
+                //Offsetting the edges to avoid z-fighting problems occuring with Unity 20.2
+                o.pos.xy *= lerp(.99, 1, unity_OrthoParams.w);
+                //Moving edges closer when using orthographic camera
+                o.pos.z *= lerp(.99, .95, unity_OrthoParams.w);
+                o.pos = mul(UNITY_MATRIX_P, o.pos);
+
                 o.color = v.color;
 
                 return o;
--- a/Content/Shader/EdgePicker.shader
+++ b/Content/Shader/EdgePicker.shader
@@ -27,6 +27,7 @@ CGPROGRAM
             #pragma vertex vert
             #pragma fragment frag
             #include "UnityCG.cginc"
+            #include "ProBuilderCG.cginc"
 
             struct appdata
             {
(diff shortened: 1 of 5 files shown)


Content/Shader/EdgePicker.shader
```diff
@@ -43,7 +43,13 @@ CGPROGRAM
             v2f vert (appdata v)
             {
                 v2f o;
-                o.pos = UnityObjectToClipPos(v.vertex);
+                o.pos = float4(UnityObjectToViewPos(v.vertex.xyz), 1);
+                //Offsetting the edges to avoid z-fighting problems occuring with Unity 20.2
+                o.pos.xy *= lerp(.99, 1, unity_OrthoParams.w);
+                //Moving edges closer when using orthographic camera
+                o.pos.z *= lerp(.99, .95, unity_OrthoParams.w);
+                o.pos = mul(UNITY_MATRIX_P, o.pos);
```
karl-, line 51: Why is z done separately here? Can this use `ProBuilderCG.UnityObjectToClipPosWithOffset` instead?

lopezt-unity: `ProBuilderCG.UnityObjectToClipPosWithOffset` was created for Metal and it does not seem to behave the same.
I moved this method to `ProBuilderCG.UnityObjectToClipPosWithOffsetMeta`l and create a 'new' `ProBuilderCG.UnityObjectToClipPosWithOffset` to handle EdgePicker as well as Face highlight the same way.
I also modified the Z to behave the same for both camera configs (ortho or not)

karl- (approved):

gytis-zasciurinskas (approved):
Tested with:
Version: 2021.2.0a7.1122
Revision: trunk 1caa39c8ddb6
Built: Wed, 24 Feb 2021 01:13:17 GMT

And with:
Version 2019.4.21f1 (b76dac84db26)
Thu, 18 Feb 2021 02:43:41 GMT
Branch: 2019.4/staging

Checked the compatibility between Legacy, URP and HDRP - everything works as expected now. Looked into the selection in Perspective, Isometric and 2D modes and it was consistent across all tests. Looks good to me!

PR #373, opened 2021-03-01
Bugfix/pb5 pass7

lopezt-unity:
### Purpose of this PR

Updating ProBuilder 5 regarding last UX pass and bugfixes.

### Links

https://docs.google.com/document/d/1rRlX73yOBH9C9F_W_dooZyGJexV8jP5_ecBSv9IXevM/edit#
**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1317773/


lopezt-unity:
@karl- and @gytis-zasciurinskas 
I opened the PR but I'm adding more corrections, I'll ping you when it'll be ready for review :)

gytis-zasciurinskas (approved):
We did a peer review with Thomas and went over all the reported issues in the doc and all of them were fixed by the PR. LGTM!:)

I have also spent some extra time going over the Undo and Preferences Reset functionality.
Undo scenarios checked:
- Creating/Deleting a Shape and PolyShape
- Resizing the Shape
- Rotating the Shape
- Changing the Shape Properties using Create Shape window

Shape Properties checked after performing Preferences Reset:
- Pivot points
- Arch settings (Thickness, Sides Count, Arch Circumference)
- Cone settings (Sides Count, Smooth)
- Cylinder settings (Sides Count, Height Cuts, Smooth)
- Door settings (Pediment Height, Side Width)
- Pipe settings (Thickness, Sides Count, Height Cuts, Smooth)
- Plane settings (Height Cuts, Width Cuts)
- Sphere settings (Subdivisions, Smooth)
- Stair settings (Steps Count, Circumference, Sides)
- Torus settings (Tube Radius, Rows, Columns, Hor. Circ., Vert. Circ., Smooth)

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -108,7 +114,8 @@ void OnEnable()
 
             m_IconContent = new GUIContent()
             {
-                image = IconUtility.GetIcon("Tools/ShapeTool/Arch"),
+                //image = IconUtility.GetIcon("Tools/ShapeTool/Arch"),
+                image = IconUtility.GetIcon("Toolbar/Panel_Shapes"),
```
karl-, line 118: remove unused code

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -406,13 +415,29 @@ void OnOverlayGUI(UObject overlayTarget, SceneView view)
...
+
         void DrawShapeGUI()
         {
             if(m_BoldCenteredStyle == null)
                 m_BoldCenteredStyle = new GUIStyle("BoldLabel") { alignment = TextAnchor.MiddleCenter };
 
             EditorGUILayout.LabelField(EditorShapeUtility.shapeTypes[s_ActiveShapeIndex.value], m_BoldCenteredStyle, GUILayout.ExpandWidth(true));
 
+            if(EditorShapeUtility.s_ResetUserPrefs.value)
+                ResetPrefs();
```
karl-, line 439: This looks pretty fragile. If there isn't already a callback when user settings are reset, we should add one and react to events rather than polling.

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -406,13 +415,29 @@ void OnOverlayGUI(UObject overlayTarget, SceneView view)
...
 
+        void ResetPrefs()
+        {
+            var type = EditorShapeUtility.availableShapeTypes[s_ActiveShapeIndex];
+            if(currentShapeInOverlay == m_LastShapeCreated)
+                m_LastShapeCreated = null;
+
+            UndoUtility.RegisterCompleteObjectUndo(currentShapeInOverlay, "Change Shape");
+            currentShapeInOverlay.SetShape(EditorShapeUtility.CreateShape(type), currentShapeInOverlay.pivotLocation);
+            SetBounds(currentShapeInOverlay.size);
```
karl-, line 426: I don't think the active shape is a setting that should respect the "reset all" action. If it's not already the case, it should be that these operational types of settings (ex, active edit mode or current shape) should be stored irrespective of the project or user level preferences.

Editor/EditorCore/EditShapeTool.cs
```diff
@@ -77,7 +82,8 @@ public override GUIContent toolbarIcon
                 if(s_IconContent == null)
                     s_IconContent = new GUIContent()
                     {
-                        image = IconUtility.GetIcon("Tools/ShapeTool/Arch"),
+                        //image = IconUtility.GetIcon("Tools/ShapeTool/Arch"),
+                        image = IconUtility.GetIcon("Toolbar/Panel_Shapes"),
```
karl-, line 86: remove unused code

Editor/EditorCore/EditorShapeUtility.cs
```diff
@@ -12,8 +13,23 @@ namespace UnityEditor.ProBuilder
 {
     internal static class EditorShapeUtility
     {
+        [UserSetting]
+        public static Pref<bool> s_ResetUserPrefs = new Pref<bool>("ShapeComponent.ResetSettings", true);
```
karl-, line 17: noted elsewhere already, but for posterity let's not rely on polling here.

Editor/EditorCore/EditorShapeUtility.cs
```diff
@@ -84,29 +100,40 @@ public static List<GUIContent[]> shapeTypesGUI
 
 
         static EditorShapeUtility()
+        {
+            ResetPrefs();
```
karl-, line 104: why are we resetting the preferences in constructor?

Editor/EditorCore/DrawShapeTool.cs
lopezt-unity: this is intentional as tools icons have been reverted to old icons. I wanted to keep the link to the newer icon version when we will update that back later.

Editor/EditorCore/DrawShapeTool.cs
lopezt-unity: Yeah I agree on this. The other way to do this would be to modify the data structure used to store the current shape prefs.
This way seems less intrusive to me and doesn't need any modification from trunk as well.

Editor/EditorCore/EditShapeTool.cs
lopezt-unity: same as above

Editor/EditorCore/EditorShapeUtility.cs
lopezt-unity: Reset Prefs is also in charge of Initializing the data structure storing the preferences if not done yet

Editor/EditorCore/DrawShapeTool.cs
lopezt-unity: Indeed, the active shape does not need to respect the resetall function. This can be removed.
However the shape settings still should on their side

Editor/EditorCore/DrawShapeTool.cs
karl-: Ok, let's not block release for this but I think we should refactor this asap

karl- (approved):

PR #376, opened 2021-03-12
Bugfixes 5x release

karl-:
### Bug Fixes

- Fixed `Draw Shape` tool showing incorrect UVs when setting shape bounding box.
- Fixed `Draw Shape` tool not clearing selection when changing the active shape type.
- Fixed `Cut Tool` error when pressing `Backspace` with no vertices placed.
- Fixed `Cut Tool` error when finishing a cut segment with less than 3 vertices.
- Fixed `Draw Shape` tool truncating shape property fields in the Scene View Overlay.

### Changes

- Moved contents of warning box in `Draw Shape` tool to tooltips.

### DIFF
--- a/Debug/Editor/Guides.cs
+++ b/Debug/Editor/Guides.cs
@@ -1,0 +1,96 @@
+using UnityEditor;
+using UnityEditor.SettingsManagement;
+using UnityEngine;
+using UnityEngine.Rendering;
+
+namespace ProBuilder.Debug.Editor
+{
+    static class GuidesSettingsProvider
+    {
+        const string k_PreferencesPath = "Preferences/Scene Guides";
+        static Settings s_Settings;
+        public static Settings settings => s_Settings ??= new Settings("com.unity.scene-guides");
+
+        [SettingsProvider]
+        static SettingsProvider CreateSettingsProvider()
+        {
+            var provider = new UserSettingsProvider(k_PreferencesPath,
+                settings,
+                new[] { typeof(GuidesSettingsProvider).Assembly });
+
+            settings.afterSettingsSaved += HandleUtility.Repaint;
+
+            return provider;
+        }
+    }
+
+    class Pref<T> : UserSetting<T>
+    {
+        public Pref(string key, T value, SettingsScope scope = SettingsScope.Project)
+            : base(GuidesSettingsProvider.settings, key, value, scope) { }
+    }
+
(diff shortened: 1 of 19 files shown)


Debug/Editor/Guides.cs
```diff
@@ -0,0 +1,99 @@
+using UnityEditor;
```
karl-, line 1: To reviewers - this is only available in dev checkouts, it is not included in the built package. It's really handy when debugging pivot issues, so I added to the repo.

Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -238,60 +239,79 @@ internal void SetBounds(Vector3 size)
...
-                    cornerPosition.y = position.y;
-                    m_Bounds.center = cornerPosition + new Vector3(size.x/2f,0, size.z/2f) + (size.y / 2f) * m_Plane.normal;
-                    m_PlaneRotation = Quaternion.LookRotation(m_PlaneForward,m_Plane.normal);
-
-                    m_BB_Origin = m_Bounds.center - m_PlaneRotation * (size / 2f);
-                    m_BB_HeightCorner = m_Bounds.center + m_PlaneRotation * (size / 2f);
-                    m_BB_OppositeCorner = m_BB_HeightCorner - m_PlaneRotation * new Vector3(0, size.y, 0);
-                    break;
-            }
+                if (HandleUtility.nearestControl == m_ControlID && previewShortcutActive)
```
antoine-unity, line 250: Not sure how true nearestControl will be depending on when this is called. If it's the end of the OnGUI it will be, otherwise a control that is closer might not have been added yet.

karl-: this only matters on repaint, so it will be valid

karl-: no wait you're right, fixing

gytis-zasciurinskas (commented):
I have tested the PR and I could no longer reproduce the listed issues. However, I have noticed one small issue. Two option fields for the Stair Shape are overlapping with their values in the Shape Creation overlay
Unity_ZIpYwZhw1h

JoelFortin (approved):
LGTM
Peer tested with Gytis and a good test pass around the fixes.

antoine-unity (approved):

PR #378, opened 2021-03-18
Fixing 'Gen Shadow Obj' tool from ProBuilder window

lopezt-unity:
### Purpose of this PR
Fixing Gen Shadow Obj tool from ProBuilder window.

The problem was due to a bad initialization of ProBuilderMesh colors table on domain reload.
In the case were colors table is null, this would result in a table of 0 elements on domain reload and create the current problem with a count difference between the vertex and the color count.
Preventing this by settings colors table to null if no elements are in the table.

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1322032/


karl- (approved):

JoelFortin (approved):
Peer-reviewed with Thomas. LGTM.

PR #379, opened 2021-03-19
Preventing creation of empty submeshes when probuilderizing 

lopezt-unity:
### Purpose of this PR

Preventing creation of empty submeshes when probuilderizing meshes with multiple materials but no submeshes.

### Links

https://fogbugz.unity3d.com/f/cases/1320936/

### DIFF
--- a/Runtime/Core/Submesh.cs
+++ b/Runtime/Core/Submesh.cs
@@ -186,6 +186,8 @@ namespace UnityEngine.ProBuilder
                 }
             }
 
+            //Removing submeshes that does not contain vertices
+            submeshes = submeshes.Where(submesh => submesh.m_Indexes.Length != 0).ToArray();
             return submeshes;
         }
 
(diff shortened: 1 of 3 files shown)


Runtime/Core/Submesh.cs
```diff
@@ -186,6 +186,8 @@ public static Submesh[] GetSubmeshes(IEnumerable<Face> faces, int submeshCount,
                 }
             }
 
+            //Removing submeshes that does not contain vertices
+            submeshes = submeshes.Where(submesh => submesh.m_Indexes.Length != 0).ToArray();
```
karl-, line 190: I expect that this will be an edge case, so if possible let's avoid allocating a second collection.

It also occurs to me that this could potentially remap face materials - if for example submesh 0 and 2 are valid, but submesh 1 is not, then this will incorrectly remap submesh 2 faces to material 1.

karl- (approved):

JoelFortin (approved):
Tested and peer-reviewed with Thomas.

PR #382, opened 2021-03-23
Re-adding menu entry for the new shape tool

lopezt-unity:
### Purpose of this PR

Re-adding menu entry for the new shape tool.
The menu entry was not present anymore, I re-added it in that PR.

### DIFF
--- a/Editor/EditorCore/EditorToolbarMenuItems.cs
+++ b/Editor/EditorCore/EditorToolbarMenuItems.cs
@@ -53,6 +53,24 @@ namespace UnityEditor.ProBuilder
 			}
 		}
 
+		[MenuItem(k_MenuPrefix + "Editors/New Shape Toggle %#k", true, PreferenceKeys.menuEditor + 1)]
+		static bool MenuVerify_NewShapeToggle()
+		{
+			var instance = EditorToolbarLoader.GetInstance<NewShapeToggle>();
+			return instance != null && instance.enabled;
+		}
+
+		[MenuItem(k_MenuPrefix + "Editors/New Shape Toggle %#k", false, PreferenceKeys.menuEditor + 1)]
+		static void MenuPerform_NewShapeToggle()
+		{
+			var instance = EditorToolbarLoader.GetInstance<NewShapeToggle>();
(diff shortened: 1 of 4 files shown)


Editor/EditorCore/EditorToolbarMenuItems.cs
```diff
@@ -53,6 +53,24 @@ static void MenuPerform_NewPolyShapeToggle()
 			}
 		}
 
+		[MenuItem(k_MenuPrefix + "Editors/New Shape Toggle %#k", true, PreferenceKeys.menuEditor + 1)]
+		static bool MenuVerify_NewShapeToggle()
+		{
+			var instance = EditorToolbarLoader.GetInstance<NewShapeToggle>();
+			return instance != null && instance.enabled;
```
karl-, line 60: does probuilder need to be open for this to work? seems like it wouldn't need to, but i'm not sure off-hand if other editor menu items also require this.

lopezt-unity: For now it doesn't, it's the same for polyshape and.... bezier shape.
Moreover I just realized that you can create a bezier shape even if experimental is off through this menu, which does seems weird to me, right?

pbmenu

Editor/MenuActions/Editors/NewBezierShape.cs
```diff
@@ -26,7 +26,7 @@ public override bool hidden
 
         public override bool enabled
         {
-            get { return true; }
+            get { return Experimental.experimentalFeaturesEnabled && ProBuilderEditor.instance != null; }
```
lopezt-unity, line 29: @karl-  I would also love to add something like :
        `protected override bool hasFileMenuEntry { get { return Experimental.experimentalFeaturesEnabled; } }`
For the Bezier shape, but that would involved to rebuild the Menu entries when switching experimental mode on/off which is there not possible (Experimental not aware of the Debug classes in which GenerateMenuItems class is)

JoelFortin (approved):
Peer-reviewed with Thomas

karl- (approved):

PR #383, opened 2021-03-24
Making experimental feature as a User Pref that can be reset

lopezt-unity:
### Purpose of this PR
`Experimental feature enabled` setting in ProBuilder preferences was not rese when resetting all ProBuilder user preferences.
This PR change that and make it a user prefs along with all others.

### Comments to Reviewers
Modification coming from Gabe's pass 8 on PB5


Editor/EditorCore/Experimental.cs
```diff
@@ -26,6 +26,11 @@ internal static bool experimentalFeaturesEnabled
         [UserSettingBlock("Experimental")]
         static void ExperimentalFeaturesSettings(string searchContext)
         {
+#if PROBUILDER_EXPERIMENTAL_FEATURES
+            //After reset ScriptingDefineSymbol is still active and needs to be deactivated
+            if(!s_experimentalFeatureEnabled.value)
+                ScriptingSymbolManager.RemoveScriptingDefine(k_ExperimentalFeaturesEnabled);
```
karl-, line 32: Is there a callback in Settings Manager for when a pref, or all prefs, are reset?

lopezt-unity: No I cheked last time and didn't see any, I might have missed that however...

JoelFortin:
The bezier shape is still in the menu after turning off the experimental feature ... and still functional. And I can confirm that the recompilation of scripts was triggered before the fix when turning the Experimental features off.

lopezt-unity:
> The bezier shape is still in the menu after turning off the experimental feature ... and still functional. And I can confirm that the recompilation of scripts was triggered before the fix when turning the Experimental features off.

@JoelFortin yes it is and it was this way before as well.
The other ongoing PR is dealing with menu items and disables the Bezier shape in the menu when experimental is disabled: https://github.com/Unity-Technologies/com.unity.probuilder/pull/382

JoelFortin (approved):
LGTM. Bezier shape tweaking made by a separated PR.

Editor/EditorCore/Experimental.cs
lopezt-unity: @karl- I just checked on the settings manager repo and it doesn't seems to have callbacks neither for property changed nor reset

Editor/EditorCore/Experimental.cs
```diff
@@ -17,31 +20,25 @@ internal static bool meshesAreAssets
 
         internal static bool experimentalFeaturesEnabled
         {
-#if PROBUILDER_EXPERIMENTAL_FEATURES
-            get { return true; }
-#else
-            get { return false; }
-#endif
+            get { return s_experimentalFeatureEnabled.value; }
```
karl-, line 23: If the project settings are removed or modified, this value may not reflect reality. That is, whether `PROBUILDER_EXPERIMENTAL_FEATURES` is defined in the compile symbols is the only accurate way to judge.

Rather than trying to keep a `Pref<bool>` in sync with this value, can we add a callback to `Settings Manager` for when a particular settings instance is mass reset and handle it manually?

karl- (approved):

PR #384, opened 2021-03-25
Adding edit shape tool icon

lopezt-unity:
### Purpose of this PR

Adding an icon for EditShape tool that is different from the normal shape tool icon.
image


karl- (approved):

JoelFortin (approved):
Peer-reviewed with Thomas

PR #386, opened 2021-03-25
Fix to avoid errors thrown in the console when entering playmode

lopezt-unity:
### Purpose of this PR

Entering playmode with a ProBuilderShape selected creates error in the Console due to Serialized Properties.
The PR fixes this to enter peacefully in playmode.

Before :
playmode-error

After fix:
playmode-errorfixed

### Links

**Fogbugz:**
**Jira:**

### Comments to Reviewers

### DIFF
--- a/Editor/EditorCore/ProBuilderShapeEditor.cs
+++ b/Editor/EditorCore/ProBuilderShapeEditor.cs
@@ -22,10 +22,59 @@ namespace UnityEditor.ProBuilder
         IMGUIContainer m_ShapeField;
 
         SerializedProperty m_ShapeProperty;
+        SerializedProperty ShapeProperty
+        {
+            get
+            {
+                if(m_ShapeProperty == null)
+                    m_ShapeProperty = serializedObject.FindProperty("m_Shape");
+                return m_ShapeProperty;
+            }
+        }
+
         SerializedProperty m_ShapePivotProperty;
+        SerializedProperty ShapePivotProperty
+        {
+            get
+            {
+                if(m_ShapePivotProperty == null)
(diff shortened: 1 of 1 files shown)


JoelFortin (approved):
Peer-reviewed with Thomas. 
Note that I wasn't able to repro it but this improvements can't hurt.
Approving.

Editor/EditorCore/ProBuilderShapeEditor.cs
```diff
@@ -61,11 +110,6 @@ bool HasMultipleShapeTypes
 
         void OnEnable()
         {
-            m_ShapeProperty = serializedObject.FindProperty("m_Shape");
-            m_ShapePivotProperty = serializedObject.FindProperty("m_PivotLocation");
-            m_ShapeSizeXProperty = serializedObject.FindProperty("m_Size.x");
-            m_ShapeSizeYProperty = serializedObject.FindProperty("m_Size.y");
-            m_ShapeSizeZProperty = serializedObject.FindProperty("m_Size.z");
```
karl-, line 68: Why is this not working specifically for `ProBuilderShape`? That's concerning, and smells like a bigger problem than just null checks.

Does this reproduce with other editors? Is it specific to `SerializeReference`?

lopezt-unity: it seems to happen only the first frame on entering playmode,not sure the problem is common I will have to confirm that as it didn't happen on Joel's PC

karl- (approved):
Approving for the purpose of landing 5.02, but let's keep the bug open and find out why this is happening.

lopezt-unity:
Not repro and code has changed since

PR #391, opened 2021-04-27
Rect Selection incorrectly selects child GameObject mesh's parts when the parent GameObject is rotated or scaled

modrimkus:
### Purpose of this PR

This PR fixes an issue where Rect Selection incorrectly selects (or selects nothing and un-focuses) mesh's Vertex/Edge/Face elements when its GameObject has a rotated and/or scaled parent.

Before fix:
PB_RotatedParent_BeforeFix
After fix:
PB_RotatedParent_AfterFix

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1324374/

### Comments to Reviewers

This issue reproduces only for `SelectionPicker` selections that have the depth test option enabled. The underlying code relied on `EmptyGameObjectWithTransform` to create matching world transform temporary game objects for depth testing. Even if `rotation` instead of `localRotation` was copied between the two, it's still not possible to always match the transform if the reference transform has parents that are rotated and have non-uniform scale. The fix changes `EmptyGameObjectWithTransform` to return a GameObject that matches local transform and then parents the temporary GameObject under the original mesh's parent - resulting in matching transforms.

### DIFF
--- a/Runtime/Core/SelectionPickerRenderer.cs
+++ b/Runtime/Core/SelectionPickerRenderer.cs
@@ -540,9 +540,6 @@ namespace UnityEngine.ProBuilder
             {
                 var pb = selection[i];
 
-                GameObject go = InternalUtility.EmptyGameObjectWithTransform(pb.transform);
-                go.name = pb.name + " (Face Depth Test)";
-
                 Mesh m = new Mesh();
                 m.vertices = pb.positionsInternal;
                 m.triangles = pb.facesInternal.SelectMany(x => x.indexesInternal).ToArray();
@@ -559,10 +556,8 @@ namespace UnityEngine.ProBuilder
 
                 m.colors32 = colors;
 
-                go.AddComponent<MeshFilter>().sharedMesh = m;
-                go.AddComponent<MeshRenderer>().sharedMaterial = BuiltinMaterials.facePickerMaterial;
-                go.hideFlags = HideFlags.HideAndDontSave;
-                go.transform.SetParent(pb.transform.parent, false);
+                GameObject go = InternalUtility.MeshGameObjectWithTransform(pb.name + " (Face Depth Test)", pb.transform, m,
+                    BuiltinMaterials.facePickerMaterial, true);
 
                 pickerObjects[i] = go;
             }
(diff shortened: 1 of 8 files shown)


karl- (approved):

Runtime/Core/SelectionPickerRenderer.cs
```diff
@@ -561,6 +561,8 @@ static ISelectionPickerRenderer pickerRenderer
 
                 go.AddComponent<MeshFilter>().sharedMesh = m;
                 go.AddComponent<MeshRenderer>().sharedMaterial = BuiltinMaterials.facePickerMaterial;
+                go.hideFlags = HideFlags.HideAndDontSave;
+                go.transform.SetParent(pb.transform.parent, false);
```
lopezt-unity, line 565: Regarding the number of time we are using a similar type of code, Could we move the following in a new method for maintenance and future changes:
```
                GameObject go = InternalUtility.EmptyGameObjectWithTransform(pb.transform);
                go.name = pb.name + "  (Vertex Billboards)";
                go.AddComponent<MeshFilter>().sharedMesh = BuildVertexMesh(pb, map, ref index);
                go.AddComponent<MeshRenderer>().sharedMaterial = BuiltinMaterials.vertexPickerMaterial;
                go.hideFlags = HideFlags.HideAndDontSave;
                go.transform.SetParent(pb.transform.parent, false);
```

Such as `InternalUtility.MakeGameObjectWithTransform(string name, Transform t, Mesh m, Material mat)` ?

Tests/Editor/Picking/RectSelection.cs
```diff
@@ -116,6 +116,41 @@ void Cleanup()
         }
     }
 
+    [Test]
+    public void PickVertices_RotatedParent_DepthTestOn()
+    {
```
lopezt-unity, line 121: Great thanks

Runtime/Core/SelectionPickerRenderer.cs
modrimkus: I agree, this should have its own method at this point. Addressed this in latest changes.

lopezt-unity (approved):

JoelFortin (approved):
LGTM, thanks for the automated test.

PR #392, opened 2021-04-29
Fix for missing Gizmos Menu items for projects that have ProBuilder installed

modrimkus:
### Purpose of this PR

 This PR fixes an issue where some Gizmos Menu items would be missing in projects that have the ProBuilder package installed.

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1332226

### Comments to Reviewers

Sequence of events causing this issue:

- User selects to install PB package or opens a project that contains PB
- Domain reload is triggered
- `GizmoSetup` (C++) receives `OnBeforeDomainReload` and clears all gizmo renderers and setups
- `AnnotationManager` (C++) receives `OnDomainReload` and sets its dirty flag
- `[InitializeOnLoad]` attributes are processed and PB `HierarchyListener`'s constructor is called which in turn calls `AnnotationUtility.GetAnnotations()`
- `AnnotationManager.GetAnotations()` internally calls `AnnotationManager.Refresh()` which, due to the dirty flag being set previously, attempts to rebuild annotations from an empty `GizmoSetup` and clears the dirty flag
- When eventually `OnDidReloadDomain` triggers, `GizmoSetup` rebuilds gizmo renderers and setups
- User opens Gizmos Menu
- `AnnotationManager.Refresh()` (C++) is called but it early exits due to the previously cleared dirty flag. This is when `AnnotationManager` would normally rebuild annotations but, due to `HierarchyListener` forcing an early `AnnotationManager.Refresh()`, the dirty flag has the wrong state. 

Proposed fix:

Delay `HierarchyListener`'s call to `AnnotationUtility.GetAnnotations()` by registering for `AssemblyReloadEvents.afterAssemblyReload` callback during `HierarchyListener`'s initialization. Using `EditorApplication.delayCall` delays it further and ensures there's no race condition between `HierarchyListener` and `GizmoSetup`'s handling of `OnDidReloadDomain`.

Tested opening existing projects that contain PB and adding/removing PB in open projects with 2019.4, 2020.3 and 2021.2.012a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
-## [Unreleased]
+
+## [5.0.3] - 2021-04-01
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+## [Unreleased]
+
+### Bug Fixes
+
+- [case: 1332226] Fixed issue where some Gizmos menu items would be missing in projects that have ProBuilder package installed.
+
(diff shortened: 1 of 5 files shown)


karl- (approved):
nice detective work 🥇

lopezt-unity (approved):
Looks good, nice catch

OndrejPetrzilka:
Nice fix! I was expecting some "initialization order issue", but wasn't able to find it.

JoelFortin (approved):
This one bothers me since a while ;)
LGTM.

PR #394, opened 2021-05-19
Fixing PBShape export using FBX Exporter

lopezt-unity:
### Purpose of this PR

Fixing Probuilder mesh throwing errors with FBX Exporter due to the presence of ProBuilderShape that was not removed.
The PR calls the `StripProBuilderScripts` class now for more consistency and the code is now also aware of the difference between "FBX Export" that does not replace the PB Object and "Convert to prefab" that creates a prefab, strips PB objects and replace the object in the scene.

FBXexportResolution

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1334017/

### Comments to Reviewers

### DIFF
--- a/Addons/Fbx.cs
+++ b/Addons/Fbx.cs
@@ -4,6 +4,7 @@ using UnityEditor;
 using System.Reflection;
 using System.Linq;
 using UnityEditor.ProBuilder;
+using UnityEditor.ProBuilder.Actions;
 
 namespace UnityEngine.ProBuilder.Addons.FBX
 {
@@ -40,13 +41,6 @@ namespace UnityEngine.ProBuilder.Addons.FBX
                 }
             }
         }
-        
-        static readonly Type[] k_ProBuilderTypes = new Type[]
-        {
-            typeof(BezierShape),
-            typeof(PolyShape),
-            typeof(Entity)
-        };
 
         static FbxOptions m_FbxOptions = new FbxOptions() {
             quads = true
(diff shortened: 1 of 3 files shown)


Addons/Fbx.cs
```diff
@@ -71,12 +65,12 @@ static void TryLoadFbxSupport()
             var getMeshForComponent = FbxExporterAssembly.GetTypes()
                .Where(t => t.BaseType == typeof(MulticastDelegate) && t.Name.StartsWith("GetMeshForComponent"))
                .First(t => t.ContainsGenericParameters);
```
karl-, line 67: We should replace the reflection here with an optional dependency on FBX Exporter

Addons/Fbx.cs
```diff
@@ -93,17 +87,17 @@ static bool GetMeshForComponent(object exporter, ProBuilderMesh pmesh, object no
...
+
+            var prefabExporterType = FbxExporterAssembly.GetType("UnityEditor.Formats.Fbx.Exporter.ConvertToPrefabSettingsSerialize");
+
+            if(exporterType == prefabExporterType)
             {
-                var component = pmesh.GetComponent(type);
-                if (component != null)
-                    Object.DestroyImmediate(component);
+                // probuilder can't handle mesh assets that may be externally reloaded, just strip pb stuff for now.
+                StripProBuilderScripts.DoStrip(pmesh);
```
karl-, line 98: nice 🥇

karl- (changes requested):

karl- (approved):

JoelFortin (approved):
LGTM. Tried Export and Convert.
Also note for @vkovec, the warning we get about the temp meta file that needs to be deleted is logged right ?

vkovec:
@JoelFortin Yes I don't think the warning was specific to Probuilder, we have made a fix for the next release

vkovec (approved):

PR #400, opened 2021-08-16
Fix CSG Crash

karl-:
### Purpose of this PR

Fixed a crash caused by *nearly* co-planar polygons recursively splitting into child nodes.

New test added.

Also adds "*.unity" files to the git lfs tracking list.

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1350635/


External/CSG/Classes/Node.cs
```diff
@@ -90,44 +89,47 @@ public void Build(List<Polygon> list)
...
-                    this.front = new Node();
+                plane.SplitPolygon(list[i], polygons, polygons, listFront, listBack);
+            
+            
+            if (listFront.Count > 0)
+            {                
+                // SplitPolygon can fail to correctly identify coplanar planes when the epsilon value is too low. When
+                // this happens, the front or back list will be filled and built into a new node recursively. This 
+                // check catches that case and sorts the front/back lists into the coplanar polygons collection.
+                if (newNode && list.SequenceEqual(listFront))
```
karl-, line 116: The `SequenceEqual` here kinda sucks, but in my testing I found that the number new node splits are generally low. Typically less than a 10 or so for each leaf. 

Given that this is an experimental feature and not really intended to be a runtime operation, I didn't spend much time investigating performance impact. In the future it may be worth hashing the list, or better yet, spending some time in `Plane.SplitPolygon` and avoiding this problem in the first place.

JoelFortin (approved):
Peer-reviewed with Karl, the main cases for the crash are fixed. It's still possible to get the crash though but should be very hard to encounter.

modrimkus (changes requested):
Maybe it's somehow just on my end, but seems like `BooleanEditor` is still referencing the previous `Boolean` static class.
Screen Shot 2021-08-19 at 1 03 46 PM

modrimkus (approved):
LGTM. Crashes are much harder to reproduce and the boolean operation results look correct.

cyanlink:
> You need to review this. Because for example this polygon leads to infinite SplitPolygon calls:
> 
> `[{"x":-17367.0039,"y":24523.3},{"x":-20042.0176,"y":21612.1973},{"x":-26959.502,"y":28434.35},{"x":-25735.1914,"y":29722.6211},{"x":-20938.9434,"y":27767.0957}]`
> 
> With an -0.2f offset.

this is because the spanning branch will interpolate the vertices, then all SequenceEqual are false, this bugfix is dunked

PR #401, opened 2021-09-09
Fixing old comments, adding doc and fixing wrong return type

lopezt-unity:
### Purpose of this PR

Updating some parts of the code regarding Amy's remarks 

### Comments to Reviewers

### DIFF
--- a/Runtime/Core/ActionResult.cs
+++ b/Runtime/Core/ActionResult.cs
@@ -1,3 +1,4 @@
+using System;
 using UnityEngine;
 
 namespace UnityEngine.ProBuilder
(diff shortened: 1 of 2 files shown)


Runtime/Core/ActionResult.cs
```diff
@@ -61,12 +62,21 @@ public ActionResult(ActionResult.Status status, string notification)
...
         }
 
+        /// <summary>
+        /// Convert the ActionResult to a boolean value, true if successful and false if not.
+        /// </summary>
+        /// <returns>True if action was successful, false otherwise.</returns>
         public bool ToBool()
         {
             return status == Status.Success;
         }
```
karl-, line 72: Why does this function exist at all? There's an implicit cast right above this

Runtime/Core/ActionResult.cs
```diff
@@ -61,12 +62,21 @@ public ActionResult(ActionResult.Status status, string notification)
...
 
-        public static bool FromBool(bool success)
+        /// <summary>
+        /// Convert a boolean value to an ActionResult, Success if true and Failure if false.
+        /// </summary>
+        /// <param name="success">The bool value to convert</param>
+        /// <returns>Success if success is true and Failure if false.</returns>
+        public static ActionResult FromBool(bool success)
         {
```
karl-, line 80: Would propose this be a constructor rather than static method.

I suppose these are already public though. So given that this is just documentation it looks OK to me, but let's deprecate these with the next major version increment.

karl- (approved):

Runtime/Core/ActionResult.cs
lopezt-unity: Yeah I guess this is an old remainder, not sure this is necessary anymore. However I guess changing this would be an API change?

Runtime/Core/ActionResult.cs
lopezt-unity: Yeah I agree on that, seems weird regarding the rest of the class to have it as a static method.
I worry that changing the return type here to correct the method is creating an API change which would require a major version bump :/

PR #405, opened 2021-10-08
Bugfix/1369443 shapetool errors domainreload

lopezt-unity:
### Purpose of this PR

The main purpose of this PR is to fix SerializedObjectNotCreatableException errors thrown in the console when using the Shape Tool and when triggering a Domain reload.

This is coming with a few modification in the way Selection is handled in the Shape tool as well as how (and when) the created shape is shown in the Inspector. For instance, the inspector is not displayed anymore when the shape is being traced in the SceneView as this could mislead the user to think that we can modifies the GameObject (pos, rotation...) while tracing the shape, which is not the case.

This PR is also fixing a few other elements :
- Polyshape MenuAction toggle  not enabled when entering in Polyshape mode
- Some API problems in the ActionResult class

### Links

https://fogbugz.unity3d.com/f/cases/1369443/

### Comments to Reviewers

Checked to be working with 2022.1, 2021.2, 2020.3 and 2019.4

### DIFF
--- a/Editor/EditorCore/DrawShapeTool.cs
+++ b/Editor/EditorCore/DrawShapeTool.cs
@@ -186,7 +186,7 @@ namespace UnityEditor.ProBuilder
                 m_CurrentState = ShapeState.ResetState();
         }
 
-        public void OnSelectionChanged()
+        void OnSelectionChanged()
         {
             if(ToolManager.IsActiveTool(this))
             {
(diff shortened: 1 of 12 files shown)


Editor/EditorCore/DrawShapeTool.cs
```diff
@@ -158,19 +179,18 @@ void OnActiveToolChanged()
...
+
         void HandleUndoRedoPerformed()
         {
             if(ToolManager.IsActiveTool(this))
                 m_CurrentState = ShapeState.ResetState();
         }
 
-        void OnSelectionChanged()
+        public void OnSelectionChanged()
         {
```
modrimkus, line 190: Curious why this was changed to public.

karl- (approved):
tested on 2022.1..0a14, wasn't able to reproduce the original bug. lgtm

JoelFortin (approved):

PR #412, opened 2022-01-21
Removing old reflection calls

lopezt-unity:
### Purpose of this PR

Removing old reflection now causing Editor crashes on Mac.
The comment was saying:
"don't let window hang around after a script reload nukes the pb_MenuAction instances"
However, this is now handled on domain reload by a delayed call to : `EditorApplication.delayCall += CloseAll;` 

### Links

**Fogbugz:** https://fogbugz.unity3d.com/f/cases/1395936/

### Comments to Reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


karl- (approved):

modrimkus (approved):

antoine-unity (approved):

JoelFortin (approved):
As long as removing this block doesn't break what was intended to handle in the first place ;)
Approving

PR #422, opened 2022-05-12
Fixed issue where vertex manipulation tools would lock grid snapping

modrimkus:
### Purpose of this PR

Fixed issue where entering any vertex manipulation tool would lock grid snapping settings (effectively leaving the user with no way to toggle it without exiting the tools).

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-3

### Comments to Reviewers

### DIFF
--- a/Editor/EditorCore/VertexManipulationTool.cs
+++ b/Editor/EditorCore/VertexManipulationTool.cs
@@ -43,6 +43,8 @@ namespace UnityEditor.ProBuilder
             get { return s_ShowHandleSettingsInScene; }
         }
 
+        public override bool gridSnapEnabled => Tools.pivotRotation == PivotRotation.Global;
+
         // Store PivotRotation so that we can detect changes and update our handles appropriately
         static PivotRotation s_PivotRotation;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix some styling issues with Overlays in 21.2
 - [case: 1350635] Fixed crash when using CSG operations.
 - [case: 1407039] Fixed stair creation tool missing the inner radius parameter.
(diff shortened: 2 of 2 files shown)


Editor/EditorCore/VertexManipulationTool.cs
```diff
@@ -43,6 +43,8 @@ internal static bool showHandleSettingsInScene
             get { return s_ShowHandleSettingsInScene; }
         }
 
+        public override bool gridSnapEnabled => Tools.pivotRotation == PivotRotation.Global;
+
```
lopezt-unity, line 47: Should that default value be changed in trunk ?

modrimkus: Yeah, I was a bit surprised why it's false by default, maybe @karl- would know the reasons behind that better? It seems that the default should be `Tools.pivotRotation == PivotRotation.Global;` instead but maybe that doesn't align well with the future work on snapping.

lopezt-unity: I was the one adding this originally and the reason I think was that most tools might not need to use grid snapping so disable is might make sense. But now I see this, I wonder if that was the good solution.

JoelFortin (approved):

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.probuilder?pullRequest=422) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.probuilder?pullRequest=422) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.probuilder?pullRequest=422) it.</sub>

PR #425, opened 2022-05-13
Fixed Collider/Trigger entity behaviours causing meshes to not render in builds

modrimkus:
### Purpose of this PR

Fixes an issue where adding and then subsequently removing a Collider or Trigger entity behaviour would have the side effect of PB mesh no longer rendering in builds. The cause of the issue was the fact that these behaviours would set the renderer's hide flags to `DontSaveInBuild` on Initialize but would not restore it to `None` on removal. 

The user's scene attached to the bug report had a number of PB meshes that did not have the behaviours attached but the renderers had their hideFlags set to `DontSaveInBuild`.

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-9

### Comments to Reviewers


Editor/EditorCore/UnityScenePostProcessor.cs
```diff
@@ -87,6 +92,9 @@ public static void OnPostprocessScene()
                 StripProBuilderScripts.DestroyProBuilderMeshAndDependencies(gameObject, mesh, true);
 
             }
+
+            foreach (var renderer in renderersToStrip)
+                Undo.DestroyObjectImmediate(renderer);
```
modrimkus, line 97: Had to move this here as otherwise I'm getting errors in the `foreach (var mesh in pbMeshes)` loop above.

karl- (approved):

lopezt-unity (approved):
LGTM

PR #431, opened 2022-06-07
St/bugfix/pbld 14

claudiaantoun:
### Purpose of this PR

This PR fixes an issue with the undo/redo operations of a newly created shape. In fact, after a redo operation, the newly created shape would either not appear in the Scene View or appear at a different position.

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-13

### Comments to Reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [case: PBLD-7] Fixed ProBuilderize action not handling redo operation correctly.
 - [case: PBLD-6] Fixed unnecessary reimport of all project textures on package install.
 - [case: PBLD-11] Fixed Poly Shape Tool missing an undo step after setting height.
+- [case: PBLD-13] Fixed undo/redo of a newly created shape.
 
 ### Changes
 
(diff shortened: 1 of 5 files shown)


lopezt-unity (approved):
LGTM, I would maybe not create a new state to do that, maybe using 
`UndoUtility.RegisterCreatedObjectUndo(tool.currentShapeInOverlay.gameObject, "Draw Shape");`
directly in the `EndState` of `ShapeState_DrawHeightShape` should work as fine and simplify the PR.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [case: PBLD-7] Fixed ProBuilderize action not handling redo operation correctly.
 - [case: PBLD-6] Fixed unnecessary reimport of all project textures on package install.
 - [case: PBLD-11] Fixed Poly Shape Tool missing an undo step after setting height.
+- [case: PBLD-13] Fixed undo/redo of a newly created shape.
```
JonMUnity, line 29: I think we could be more descriptive of the original bug here. 
I took a look at the bug description on Jira. What do you think of this suggestion? Does it work? 

```suggestion
- [case: PBLD-13] Fixed a bug where a newly created shape would not be redrawn with Redo. 
```

JonMUnity (approved):

karl- (approved):

PR #440, opened 2022-09-07
Preserve existing mesh data in ProBuilderMesh.MakeUnique.

DaZombieKiller:
### Purpose of this PR
This fixes a rather strange issue that we encountered at New Blood while using ProBuilder in Gloomwood.

We would frequently find that after baking lightmaps, a random selection of ProBuilder meshes would have their lightmap UVs wiped. Notably, this UV issue would not manifest until either the objects in question were selected in the editor, or a build was produced.

After some investigation, we managed to track the issue down to the `MakeUnique` function on `ProBuilderMesh`, which seems to re-generate the mesh even if one already exists, thus discarding the lightmap UV channel and anything else not stored directly on the component.

It's not clear to me why these objects had `MakeUnique` called on them after a lightmap bake, so there is likely another issue at play _in addition_ to this one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - [case: PBLD-32] Fixed `New Shape` start location being incorrect after using right mouse button.
 - [case: PBLD-19] Fixed shape creation when the camera perspective is set to Top.
+- `ProBuilderMesh.MakeUnique` now preserves existing mesh data when possible.
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.probuilder?pullRequest=440) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.probuilder?pullRequest=440) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.probuilder?pullRequest=440) it.</sub>

karl- (approved):

karl- (approved):

karl-:
Incorporated into https://github.com/Unity-Technologies/com.unity.probuilder/pull/453, as CI doesn't allow forks to run tests.

PR #447, opened 2022-11-09
Fix for no UV distribution metric recalculation after mesh optimization step

modrimkus-unity:
### Purpose of this PR

PR fixes an issue where the UV distribution metric was not recalculated after rebuilding a mesh. This would have the side effect of incorrect mip maps being loaded.

Tested for possible performance implications: used a 5k vertex cube and sampled the `EditorMeshUtility. Optimize` function. With `RecalculateUVDistributionMetric` or without  the measurements fluctuated between 55ms and 70ms, therefore if there is any impact, it's pretty negligible and probably around additional 5-10ms per 5K vertices.

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-41
**Slack:** https://unity.slack.com/archives/C7HJDB0A0/p1666627588183009

### Comments to Reviewers

### DIFF
--- a/Editor/EditorCore/EditorMeshUtility.cs
+++ b/Editor/EditorCore/EditorMeshUtility.cs
@@ -49,7 +49,9 @@ namespace UnityEditor.ProBuilder
                 if (umesh.GetTopology(i) != MeshTopology.Triangles)
                     skipMeshProcessing = true;
 
+#pragma warning disable 0219
             bool uv2Built = false;
+#pragma warning restore 0219
             if (!skipMeshProcessing)
             {
                 bool autoLightmap = Lightmapping.autoUnwrapLightmapUV;
@@ -111,9 +113,11 @@ namespace UnityEditor.ProBuilder
             if (Experimental.meshesAreAssets)
                 TryCacheMesh(mesh);
 
+#if UNITY_2020_2_OR_NEWER
             umesh.RecalculateUVDistributionMetric(0);
             if (uv2Built)
                 umesh.RecalculateUVDistributionMetric(1);
+#endif
 
             UnityEditor.EditorUtility.SetDirty(mesh);
         }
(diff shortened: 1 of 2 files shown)


Editor/EditorCore/EditorMeshUtility.cs
```diff
@@ -108,6 +111,10 @@ public static void Optimize(this ProBuilderMesh mesh, bool generateLightmapUVs =
             if (Experimental.meshesAreAssets)
                 TryCacheMesh(mesh);
 
+            umesh.RecalculateUVDistributionMetric(0);
+            if (uv2Built)
+                umesh.RecalculateUVDistributionMetric(1);
```
karl-, line 116: If I understand this correctly, the minimum UV area to consider value is relative to the texture size, where one texel is equal to `1 / (texture width, height)`. So it looks like this would be effectively ignoring most of the UVs on the mesh.

The default is 1e-9f, why not leave it at that?

Also I'm not sure I understand why it is first called with a parameter of `0`, then if UV2 is built, again with `1`.

karl- (changes requested):

Editor/EditorCore/EditorMeshUtility.cs
modrimkus-unity: It actually has two signatures, I'm using this one: https://docs.unity3d.com/ScriptReference/Mesh.RecalculateUVDistributionMetric.html 
The first param is the UV set, the second is the threshold, which I left to default.

modrimkus-unity: Addressed compatibility with older editors in [08af8da](https://github.com/Unity-Technologies/com.unity.probuilder/pull/447/commits/08af8daaec54014b7be0b9d920de8265bf24f60a).

karl-: Cool, I didn't notice the overload. Looks good to me.

PR #449, opened 2023-01-11
[PBLD-48] Change Shape's min size

lopezt-unity:
### Purpose of this PR

Fixed a bug where the minimum size for a shape was not taking into account the snap and grid sizes.
The min size was hard coded and not taking the snapping or grid sizes into account, which it does now if these sizes are smaller than the default one (0.1f);

Before:
bug-pb-0 01

After:
fix-pb-0 01

### Links

**Jira:** https://jira.unity3d.com/browse/UUM-19007

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
-- [case: PBLD-48] Fixed a bug where the minimum size for a shape was not taking into account the snap and grid sizes.
+- [case: PBLD-48] Fixed a bug where the minimum size of a shape did not take into account snap and grid sizes.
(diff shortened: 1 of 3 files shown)


karl- (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- [case: PBLD-48] Fixed a bug where the minimum size for a shape was not taking into account the snap and grid sizes.
```
JonMUnity, line 12:
```suggestion
- [case: PBLD-48] Fixed a bug where the minimum size of a shape did not take into account snap and grid sizes.
```

modrimkus-unity (approved):

PR #463, opened 2023-06-05
fix arch shape not allowing 2 sides

karl-:
### Purpose of this PR

makes shape behaviour consistent with documentation, and documentation consistent with images.


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Moved cut tool cursor icon from runtime to editor resources.
 - Added `ProBuilderMesh.MakeUnique` to ensure `Mesh` asset is distinct.
+- Arch shape now allows 2 sides.
```
JonMUnity, line 26: Should be in `### Fixed`.

Documentation~/Arch.md
```diff
@@ -16,10 +16,10 @@ When you customize an arch shape, keep in mind that when you change the thicknes
...
+|:-- |:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| __Thickness__ | Set the thickness of the arch in meters. The larger the thickness, the smaller the opening becomes. The default value is 0.1. The minimum value is 0.01.                                                                                                                                                                                 |
+| __Sides Count__ | Set the number of sides for the arch. The more sides you use (relative to the size of the __Radius__), the smoother the arch becomes, so the closer you get to a semi-circle. Conversely, if you this value to 2, the arch turns into a peaked roof, regardless of the radius. The default value is 5. Valid values range from 2 to 200. |
```
JonMUnity, line 22:
```suggestion
| __Sides Count__ | Set the number of sides for the arch. The more sides you use (relative to the size of the __Radius__), the smoother the arch becomes, so the closer you get to a semi-circle. Conversely, if you set this value to 2, the arch turns into a peaked roof, regardless of the radius. The default value is 5. Valid values range from 2 to 200. |
```

CHANGELOG.md
JonMUnity:
```suggestion
- The Arch shape can now have 2 sides. .
```

PR #470, opened 2023-07-18
Fix poly shape regression - not being able to set an extrusion point.

claudiaantoun:
### Purpose of this PR

This PR fixes a regression in the poly shape tool, with the user not being able to set an extrusion point, that I caused when fixing these bugs (https://jira.unity3d.com/browse/PBLD-52 and https://jira.unity3d.com/browse/PBLD-70).

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-78

### DIFF
--- a/Runtime/Core/ProBuilderMeshFunction.cs
+++ b/Runtime/Core/ProBuilderMeshFunction.cs
@@ -368,6 +368,18 @@ namespace UnityEngine.ProBuilder
 
             mesh.subMeshCount = submeshes.Length;
 
+            if (mesh.subMeshCount == 0)
+            {
+                EnsureMeshFilterIsAssigned();
+
+                if (usedInParticleSystem)
+                    MeshUtility.RestoreParticleSystem(this);
+
+                IncrementVersionIndex();
+
+                return;
+            }
+
             var currentSubmeshIndex = 0;
             var shouldReassignMaterials = false;
             for (int i = 0; i < mesh.subMeshCount; i++)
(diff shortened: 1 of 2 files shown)


Runtime/Core/ProBuilderMeshFunction.cs
```diff
@@ -368,6 +368,18 @@ public void ToMesh(MeshTopology preferredTopology = MeshTopology.Triangles)
...
+            {
+                EnsureMeshFilterIsAssigned();
+
+                if (usedInParticleSystem)
+                    MeshUtility.RestoreParticleSystem(this);
+
+                IncrementVersionIndex();
+
+                return;
+            }
```
modrimkus-unity, line 381: Would be nice if the code duplication could be avoided (i.e. compared to the very bottom code of this function). Also a comment might be nice for future changes that would explain why the materials/submesh code below can't be executed/is problematic when subMeshCount == 0.

modrimkus-unity (approved):
Approving but left a comment for discussion.

karl- (approved):
+1 to Mod's comment. Otherwise no issues.

Runtime/Core/ProBuilderMeshFunction.cs
claudiaantoun: Good point, thanks for the feedback! Fixed in [4e34729](https://github.com/Unity-Technologies/com.unity.probuilder/pull/470/commits/4e34729e31f3f46b08484045802645828293c616)

PR #475, opened 2023-08-03
API Update for Analytics

lopezt-unity:
### Purpose of this PR

API has been changed in unity 2023.2.0a22 and after for analytics causing warning messages in the console when using ProBuilder. 
This PR updates the analytics to use the new API for 2023.2 and after.

Based on this doc : https://docs.editor-data.unity3d.com/Contribute/EditorAnalytics/cs_guide/

### Links

**Jira:** https://jira.unity3d.com/browse/STO-2988

### Comments to Reviewers

### DIFF
--- a/Editor/EditorCore/ProBuilderAnalytics.cs
+++ b/Editor/EditorCore/ProBuilderAnalytics.cs
@@ -11,17 +11,141 @@
 //#define PB_ANALYTICS_DONTSEND
 
 using System;
-using System.Collections.Generic;
-using System.Linq;
 using System.IO;
-using UnityEditor;
 using UnityEditorInternal;
-using UnityEngine;
 using UnityEngine.Analytics;
 using UnityEngine.ProBuilder;
+#if !UNITY_2023_3_OR_NEWER
+using System.Linq;
+using System.IO;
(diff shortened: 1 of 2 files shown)


Editor/EditorCore/ProBuilderAnalytics.cs
```diff
@@ -11,29 +11,55 @@
 //#define PB_ANALYTICS_DONTSEND
```
karl-, line 11: Can we add `#if !DEBUG` here to exclude developer builds?

karl- (approved):
non-blocking question, otherwise looks good

Editor/EditorCore/ProBuilderAnalytics.cs
lopezt-unity: I added that in [40fa1de](https://github.com/Unity-Technologies/com.unity.probuilder/pull/475/commits/40fa1de3f1d9570b95e15929a997332213ac3a60) but at the Send event level so that we can still log analytics in a file to test but they are not sent to the analytics for real :D
Does that sound ok ?

modrimkus-unity (approved):

PR #480, opened 2023-08-31
[PBLD-87] Fixing infinite recursion

lopezt-unity:
### Purpose of this PR

Fixed a bug where an infinite recursion loop could crash the editor when using the Boolean tool.

The problem was caused by the creation of a new polygon every single time using:
`{front/back}.Add(new Polygon(f, polygon.material));`

In some cases, the created polygon was slightly different from the original one, causing a stoppin condition to fail in Node.Build  (line 116 or line 124), causing another call to Build() with the same polygin vertices, causing the same issue on the following loop.
`if (newNode && list.SequenceEqual(listBack/Front))`

Checking the vertex equality before creating the polygon fixes that, then the original polygon is used, instead of a new one.

Also adding a small optimization, checking that the new added polygon is valid before adding it to the list.

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-87

### Comments to Reviewers

### DIFF
--- a/External/CSG/Classes/Plane.cs
+++ b/External/CSG/Classes/Plane.cs
@@ -1,5 +1,6 @@
 using UnityEngine;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace UnityEngine.ProBuilder.Csg
 {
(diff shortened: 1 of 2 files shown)


antoine-unity (approved):

External/CSG/Classes/Plane.cs
```diff
@@ -126,12 +127,26 @@ public void SplitPolygon(Polygon polygon, List<Polygon> coplanarFront, List<Poly
 
                     if (f.Count >= 3)
                     {
-                        front.Add(new Polygon(f, polygon.material));
+                        if (f.SequenceEqual(polygon.vertices))
+                            front.Add(polygon);
+                        else
+                        {
+                            var p = new Polygon(f, polygon.material);
+                            if (p.plane.Valid())
```
karl-, line 135: in what case would the new polygon not be valid?

karl- (approved):
would be nice if we could avoid linq here. i'm guessing that these polygon vertex lists are generally quite small? would it be a perf improvement to test dot product before doing a sequence compare?

Not blocking feedback, however. For the sake of this utility I think stability is more important than performance. Approved :+1:

PR #491, opened 2023-11-21
[PBLD-88] Fixing ProBuilder Window not refreshing

lopezt-unity:
### Purpose of this PR

Fixed a bug that prevented the ProBuilder window to update its content after changing its docking position.

The problem was caused by the fact that the events when registered in the constructor and unregistered when docking was changed from the original position, and then never registered again.
Moving the event registration to the AttachToPanelEvent event fixes this problem.

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-88

### Comments to Reviewers


modrimkus-unity (approved):
LGTM!

karl- (approved):
Ugh silly mistake. Thank you Thomas 🥇

PR #495, opened 2023-12-12
Preview improvement

lopezt-unity:
### Purpose of this PR

Regarding last week reviews, the main blocker for PB6 was the preview button, this was also casing inconsistencies between the parameter settings and the applied values...
This PR fixes that by improving live preview.
Also adding a parameter in case of dealing with large meshes and in that case only updating preview on mouse release/field edition.

Unity_cOs9AfG39f

**Jira:**

### Comments to Reviewers

### DIFF
--- a/Debug/Editor/GenerateMenuItems.cs
+++ b/Debug/Editor/GenerateMenuItems.cs
@@ -208,8 +208,11 @@ namespace UnityEditor.ProBuilder
                     sb.AppendLine($"\t\t\t\tEditorAction.Start(new MenuActionSettings(instance,{data.hasPreview.ToString().ToLower()}));");
                 }
                 else // Last case, just perform the action if this is an instant action with no options
+                {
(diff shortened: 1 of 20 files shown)


karl- (approved):

lopezt-unity:
It's a preference indeed, but as @gabrielw-us told me when I showed that to him is that depending on the context and the current action (and selection), you might want to change that often.

> Nice improvement. Minor suggestions that do not need to block :
> 
> 1. Does "Live Preview" toggle need to be in the Overlay? My intuition would be that this is a preference.
> 2. "Validate" isn't the correct word for what it does. "Confirm" would be better  ( cc @JonMUnity )

modrimkus-unity (approved):
Looking great! I agree with @gabrielw-us  that the live preview setting should be in the overlay (and that was also my first reaction when @lopezt-unity showed the improvement). One argument for it is that once the user will learn that it can be sluggish if the selection is large for example, it's better to have control for on/off always on hand rather than have to go to prefs.

PR #500, opened 2024-02-06
One shader to rule them all (AKA simplifying render pipeline workflows).

lopezt-unity:
### Purpose of this PR

Simplifying the shaders imports for HDRP and SRP. 
Adding the Shader Graph dependency to be able to support multi RP shaders. 

Based on #499 

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-102

### Comments to Reviewers

[List known issues, planned work, provide any extra context for your code.]

PR #501, opened 2024-02-07
Cleaning ProBuilder 6 icons

lopezt-unity:
### Purpose of this PR

Removed a lot of menu actions icons that are not used anymore.

However, I let the paths in the associated Menu Actions to the former value. It seems simpler to keep these and not use them if it needs to be reverted later at some point. The icon path could also be removed from the menu actions if that makes more sense regarding the current state as well.

Let me know for that.

I also removed a script that didn't seem to be used anymore and for a while already: `Interaction/ToggleHandlePivotPoint.cs`.
It seems it is now handled by the pivot position tool settings.

Based on #500 

### Links

**Jira:** https://jira.unity3d.com/browse/STO-3174

### Comments to Reviewers


claudiaantoun (approved):
LGTM! Like we discussed, I think if it doesn't make a different when reverting this change, we should remove the icon paths as well.

claudiaantoun:
I got these errors when I switched to the branch to test.
image

lopezt-unity:
This is a known issue and has been fixed in trunk recently :)


> I got these errors when I switched to the branch to test. I think this issue was resolved in another of your PRs, right?
> 
> image

karl- (approved):

PR #509, opened 2024-05-07
[Backport][PBLD-110] Updating prefab instances

lopezt-unity:
### Purpose of this PR

[Backport from 6.0.x] https://github.com/Unity-Technologies/com.unity.probuilder/pull/508

When modifying a pb mesh in a prefab and applying the ovrerrides, some prefabs where not updated.
This is beccause in that case the versionIndex could be set to 0 and as the ProBuilderMesh.meshSyncState says :

                // if the local version index is uninitialized, that means no edits have occurred since loading this
                // mesh. it could mean a new mesh was created and ToMesh has not been called yet, or that this mesh was
                // copy/pasted from an existing instance.
                // in the latter case, it is handled by listening to the ObjectChangeEvent stream in HierarchyListener.
                // this function only cares about the case where modifications have been made during the current session.
This is also the case when the mesh is part of a prefab and as not been modified, but it should be updated as well when the instance is updated.

To solve this, we revert to the former solution where we were forcing the mesh to rebuild.

Also fixing the tests, the previous version of the test was not rebuilding the mesh so the compared data was not correct and the test was wrongly successful. The PR fixes that as well, and adds some extra setup to insure a new scene is created.

Previous behavior:
Unity_OsQuGzq5VS

Fixed behavior :
Unity_FPX5FpSThx

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- [PBLD-110] Fixed a bug where the prefab instances of ProBuilder meshes where not updating after applying all the overrides.
 - [WEBDOCS-1036] Fixed a documentation generation problem where a setting in the filter.yml file was preventing the ProBuilder API documentation from being generated.
 
 ## [5.2.2] - 2023-11-27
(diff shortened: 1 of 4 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.probuilder?pullRequest=509) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.probuilder?pullRequest=509) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.probuilder?pullRequest=509) it.</sub>

modrimkus-unity (approved):

PR #531, opened 2024-07-15
Fixed visual glitches when drawing PB shapes on macOS

modrimkus-unity:
### Purpose of this PR

PR fixes an issue where drawing the height during PB shape creation would cause visual glitches on macOS.

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-135

### Comments to Reviewers


modrimkus-unity:
@lopezt-unity I did not notice issues with other tools, at least with the most used ones. There could still possibly be a misplaced draw somewhere but we'll fix it when it surfaces.

PR #534, opened 2024-07-17
Fix PolyShape tool warning when Resident Drawer enabled

modrimkus-unity:
### Purpose of this PR

PR fixes an issue where warnings would be thrown or some parts of PolyShapes would render invisible when Resident Drawer is on.

There's too parts to the fix:
1) `EditorUtility.InitObject` no longer results configurations where there's zero submeshes but one or more materials - this would cause Resident Drawer to throw errors.
2) Material is more aggressively refreshed as otherwise geometry added after setting the material would render as if there's no material set - probably has to do with how Resident Drawer batches geometry.

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-115

### Comments to Reviewers

### DIFF
--- a/Content/Resources/Materials/ProBuilderDefault.mat
+++ b/Content/Resources/Materials/ProBuilderDefault.mat
@@ -45,7 +45,7 @@ MonoBehaviour:
 --- !u!21 &2100000
 Material:
   serializedVersion: 8
-  m_ObjectHideFlags: 0
+  m_ObjectHideFlags: 32
   m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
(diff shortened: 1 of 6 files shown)


Content/Resources/Materials/ProBuilderDefault.mat
lopezt-unity, line 1: What are the changes on the material ?

lopezt-unity (approved):
Just a question regarding the changes on the material itself

Content/Resources/Materials/ProBuilderDefault.mat
modrimkus-unity: You're too quick! yeah pushed that by accident and only noticed once I opened the PR. Reverted :)

claudiaantoun (approved):

PR #543, opened 2024-07-31
Fixed issue where duplicating pb mesh would dirty all scenes

modrimkus-unity:
### Purpose of this PR

PR fixes a regression where duplicating a PB mesh would dirty all loaded scenes. 

This was caused by the usage of `Undo.RegisterCreatedObjectUndo` which has a behaviour of dirtying all scenes if it's a non-GO floating object (the engine does not know which scene the object belongs to therefore dirties all possible ones).  This is also confirmed by an old Slack thread: https://unity.slack.com/archives/C06TQ9LFP/p1584975717283900

The fix instead uses the same command as is used when creating a new PB shape `Undo.RegisterCompleteObjectUndo` - this fully records the duplicated GO while also recording the creation of a new mesh via reference. This does not have the unwanted side-effect of dirtying all scenes.

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-140

### Comments to Reviewers


Tests/Framework/TestUtility.cs
```diff
@@ -43,11 +43,11 @@ public virtual void OneTimeTearDown()
             AssetDatabase.Refresh();
         }
 
-        public Scene OpenScene(string path)
+        public Scene OpenScene(string path, OpenSceneMode openMode = OpenSceneMode.Single)
         {
             var tempPath = $"{tempDirectory}/{Path.GetFileName(path)}";
             AssetDatabase.CopyAsset(path, tempPath);
-            return EditorSceneManager.OpenScene(tempPath);
+            return EditorSceneManager.OpenScene(tempPath, openMode);
```
lopezt-unity, line 50: Side note for the backport on PB5.2 later -> This method is only available after Unity 2022.3 and later and PB5 is verified for 2019.4 :)

lopezt-unity (approved):

Tests/Framework/TestUtility.cs
modrimkus-unity: Any signature of ` EditorSceneManager.OpenScene` ? Are you sure? https://docs.unity3d.com/2019.4/Documentation/ScriptReference/SceneManagement.EditorSceneManager.OpenScene.html

claudiaantoun (approved):

PR #578, opened 2025-01-08
[PBLD192] Fix for vertex snapping to self-object vertices 

lopezt-unity:
### Purpose of this PR

This PR fixes the Move tool behavior when using Vertex snapping.

The user now has a setting to activate/de-activate in order to enable vertex snapping on the object itself.
This is a settings in the tool settings toolbar, also assignable using a shortcut.

snaptoself-fix

Leaving that as a settings as using vertex snapping on really complex objects, it might not be desired by users to snap on the object itself :P 
https://jira.unity3d.com/browse/PBLD-192

@gabrielw-us : I would need a new icon for that settings in the toolbar

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-192

### Comments to Reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,23 +7,29 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased 
 
+### Changes
+
+- Improved DrawShape tool performance when you place the first vertex.
+
 ### Fixed
 
-- [PBLD-192] Fixed an issue where vertices were not able to snap on vertices from the same mesh.
-- [PBLD-180] Fixed an issue in the Material Editor where an exception was thrown because it attempted to retrieve `MenuItem` shortcuts before the main menu was fully loaded.
-- Fixed a bug where Undo was not working correctly with Polyshape creation tool. 
-- [PBLD-196] Fixed a bug where Polyshape creation tool was not working with orthographic camera.
+- [PBLD-192] Fixed a bug where vertices were not able to snap on other vertices from the same mesh.
+- [PBLD-189] Fixed a bug where using the auto-stitch functionality of the UV Editor would not work properly on MacOS.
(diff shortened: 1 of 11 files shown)


modrimkus-unity (approved):
LGTM. Additionally did some sanity manual testing - all good.

katherinesheridan1 (approved):
LGTM !!

gabrielw-us (commented):
This looks good functionally, but I'm not sure about having a "snap to self" toggle. I'd prefer to make 2 parts:
1. A PR that fixes the bug (snapping to self should work)
2. A PR where we can consider new features like a toggle for this and other snapping options too perhaps. Mainly because we'll need at least one new icon, and that is usually a few weeks time for Deanna from the request date. And, honestly, I'm still just not seeing the need for this as a feature.

FWIW, I did check with the original code author on this area (Karl), and he confirmed this was just a bug, not intended behavior.

lopezt-unity:
Let's do 1. then. Indeed, making it as an option would be feature work if we move that to later, but that does not seem a priority at the moment and there is no user request in that direction as well. So I'll removed all the extras and focus on the main fix then :) 

> This looks good functionally, but I'm not sure about having a "snap to self" toggle. I'd prefer to make 2 parts:
> 
> 1. A PR that fixes the bug (snapping to self should work)
> 2. A PR where we can consider new features like a toggle for this and other snapping options too perhaps. Mainly because we'll need at least one new icon, and that is usually a few weeks time for Deanna from the request date. And, honestly, I'm still just not seeing the need for this as a feature.
> 
> FWIW, I did check with the original code author on this area (Karl), and he confirmed this was just a bug, not intended behavior.

lopezt-unity:
@gabrielw-us : REmoved the property in [31b92b7](https://github.com/Unity-Technologies/com.unity.probuilder/pull/578/commits/31b92b75ba06f55b708a27082fd94297d164058e) :)

PR #582, opened 2025-01-23
[PBLD-189][PB5-Backport] Fixed auto-stitch shortcuts 

lopezt-unity:
### Purpose of this PR
Fixed a bug where using the auto-stitch functionality of the UV Editor would not work properly on MacOS.
Adding the 2 following auto-stitch functions to the shortcut manager and removing them from custom code handling:

- Auto-stitch UV
- Copy UV Settings

@modrimkus-unity , can you test:
- opening the ScreenshotMenu from the button (I changed UNITY_STANDALON to EDITOR), the menu should now open
- Auto-stich shortcut for MacOS
- CopyUVSettings (Cmd + Shift shortcut should assign the UV of the selected element to the clicked one)

### Links
Jira: https://jira.unity3d.com/browse/PBLD-189

### Comments to Reviewers

### DIFF
--- a/Editor/EditorCore/UVEditor.cs
+++ b/Editor/EditorCore/UVEditor.cs
@@ -716,7 +716,7 @@ namespace UnityEditor.ProBuilder
             }
 #if UNITY_EDITOR_OSX
             // PBLD-189: Auto-stitch has incorrect and inconvenient shortcut on MacOS
-            else if (e.modifiers == (EventModifiers.Command | EventModifiers.Shift | EventModifiers.Alt))
+            else if (e.modifiers == (EventModifiers.Control | EventModifiers.Command |  EventModifiers.Shift))
 #else
             else if (e.modifiers == EventModifiers.Control)
 #endif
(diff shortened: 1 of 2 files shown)


Editor/EditorCore/UVEditor.cs
```diff
@@ -714,7 +714,12 @@ internal bool ClickShortcutCheck(ProBuilderMesh pb, Face selectedFace)
             {
                 return CopyFaceUVSettings(pb, selectedFace);
             }
+#if UNITY_EDITOR_OSX
+            // PBLD-189: Auto-stitch has incorrect and inconvenient shortcut on MacOS
+            else if (em == (EventModifiers.Command | EventModifiers.Shift | EventModifiers.Alt))
```
modrimkus-unity, line 719: Does not compile currently:

```suggestion
            else if (e.modifiers == (EventModifiers.Command | EventModifiers.Shift | EventModifiers.Alt))
```

modrimkus-unity: `Cmd + Shift + Alt` seems to not come through into `IsCopyUVSettingsModifiers` unfortunately on 2022.3 - seems that the view tool activates instead. On the otherhand, `Ctl + Cmd + Shift` does work (ctrl is valid key on macOS btw and doesn't trigger the menu if shift is involved as well).

modrimkus-unity (changes requested):

modrimkus-unity (approved):
Updated that shortcut to Control + Command + Shift + M0 as Alt seems to be triggering view tool (probably due to it not going through SM or U6 changes in SM).

Also, all of the 3 bullet points mentioned in PR are working fine.

PR #588, opened 2025-02-18
[PBLD-199] NullReferenceException when using the Cut Tool

katherinesheridan1:
### Purpose of this PR

This PR fixes an issue where if, while using the Cut Tool, two vertices were placed on the same edge, a null reference exception would be thrown in the console. This was due to the face that the tool was attempting to create a face where it should not.

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-199

### Comments to Reviewers
None.

### DIFF
--- a/Editor/EditorCore/CutTool.cs
+++ b/Editor/EditorCore/CutTool.cs
@@ -946,12 +946,16 @@ namespace UnityEditor.ProBuilder
 
                      if(index >= 0)
                      {
-                         List<Face> toDelete;
-                         Face newFace = ComputeFaceClosure(polygon, index, cutVertexSharedIndexes, out toDelete);
-                         newFace.submeshIndex = m_TargetFace.submeshIndex;
+                         // In the case of only 2 distinct, a face should not be added.
+                         if (polygon.Distinct().Count() != 2)
+                         {
+                             List<Face> toDelete;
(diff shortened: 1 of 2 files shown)


Editor/EditorCore/CutTool.cs
```diff
@@ -946,12 +946,16 @@ List<Face> ComputeNewFaces(Face face, IList<int> cutVertexIndexes)
...
                      if(index >= 0)
                      {
-                         List<Face> toDelete;
-                         Face newFace = ComputeFaceClosure(polygon, index, cutVertexSharedIndexes, out toDelete);
-                         newFace.submeshIndex = m_TargetFace.submeshIndex;
-
-                         newFaces.Add(newFace);
-                         facesToDelete.AddRange(toDelete);
+                         // In the case of only 2 distinct, a face should not be added.
+                         if (polygon.Distinct().Count() != 2)
```
lopezt-unity, line 950: I think, that should be just:
```suggestion
                         if (polygon.count != 2)
```
Reason is we should minimize the use of Linq, and also they should already be distinct :D

katherinesheridan1: Thank you !

modrimkus-unity (approved):
LGTM, seems to fix the reported error 👍

lopezt-unity (approved):
LGTM

PR #590, opened 2025-02-19
Change PB hierarchy menu's "baked" shapes to parametric shapes

modrimkus-unity:
### Purpose of this PR

PR fixes an issue where shapes created using PB's hierarchy menu would result in non-procedural/baked PB meshes. The fix makes it so that all these shapes are now editable `ProBuilderShapes` and their pivot, UVs and general topology match 1:1 to what the baked shapes had. 

Below are pairs of the prior regular PB hierarchy mesh vs new `ProBuilderShape` editable result:

Screenshot 2025-02-19 at 5 16 44 PM

In addition, and this is after confirming with Gabe on a call, the `ProBuilderShape` cube's UVing is changed to Auto, Top Left anchor so that a unit size cube always has practical UVing (instead of the cross at its center) when created both via the Create Shape tool and hierarchy menu. 

Before:
https://github.com/user-attachments/assets/1e72e9b9-141d-46a8-b13b-eb5ba0d9b036

After:
https://github.com/user-attachments/assets/43bd942d-3637-4f26-a54d-1ab7ca0f1dbd

### Links

**Jira:** https://jira.unity3d.com/browse/PBLD-202

### Comments to Reviewers


Tests/Editor/Editor/UVEditorTest.cs
```diff
@@ -91,10 +94,11 @@ public void Manual_PlanarProjection()
             Assert.That(f.manualUV, Is.EqualTo(true));
         }
 
-        //Modify those faces
+		//Confirm that UV bounds origin are not at the LowerLeft corner
         Vector2 minimalUV = UVEditor.instance.UVSelectionMinimalUV();
```
modrimkus-unity, line 98:
```suggestion
	//Confirm that UV bounds origin are not at the LowerLeft corner
        Vector2 minimalUV = UVEditor.instance.UVSelectionMinimalUV();
```

Tests/Editor/Editor/UVEditorTest.cs
```diff
@@ -18,6 +18,10 @@ public void Setup()
     {
         m_cube = ShapeFactory.Instantiate<Cube>();
         EditorUtility.InitObject(m_cube);
+		// Unsure UV bounds origin is not at (0,0) lower left
```
modrimkus-unity, line 21:
```suggestion
	// Unsure UV bounds origin is not at (0,0) lower left
```

Runtime/Shapes/Cube.cs
```diff
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
```
modrimkus-unity, line 1:
```suggestion
```

karl-:
In my opinion this should be optional behaviour. I regularly use the menu items to create shapes specifically to avoid the extraneous procedural component.

Runtime/Shapes/Shape.cs
```diff
@@ -35,6 +35,8 @@ public virtual Bounds UpdateBounds(ProBuilderMesh mesh, Vector3 size, Quaternion
         /// </summary>
         /// <param name="shape">The <see cref="Shape" /> to copy property values from.</param>
         public abstract void CopyShape(Shape shape);
+
+        internal abstract void SetParametersToBuiltInShape();
```
lopezt-unity, line 39: Just a detail here, I would maybe think that `ResetDefaultShapeParameters` is a little more explicit here ?

lopezt-unity: Whatever, ignore this

lopezt-unity (approved):
LGTM

katherinesheridan1 (approved):
Looks great !

PR #591, opened 2025-02-21
[PBLD-199] [PB 5.2 Backport] NullReferenceException when using the Cut Tool

katherinesheridan1:
### Purpose of this PR

This PR fixes an issue where if, while using the Cut Tool, two vertices were placed on the same edge, a null reference exception would be thrown in the console. This was due to the face that the tool was attempting to create a face where it should not.

**Jira:** https://jira.unity3d.com/browse/PBLD-199

### Comments to Reviewers

None.


modrimkus-unity (approved):
LGTM!

PR #604, opened 2025-04-18
correct shader property type access to use non-obsolete property

jonbro:

### Purpose of this PR

Update Access to the property type to non-obsolete call.

### Links

jira: https://jira.unity3d.com/browse/PBLD-228

### Comments to Reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changes
 
-- [PBLD-228] Updated function to access shader property types
+- [PBLD-228] Updated function to access shader property types.
 
 ### Fixed
 
-- [PBLD-224] Fixed rect selection in HDRP
+- [PBLD-224] Fixed rect selection in HDRP.
 
 ## [6.0.5] - 2025-03-11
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changes
 
+- [PBLD-228] Updated function to access shader property types
+
 ### Fixed
 
 - [PBLD-224] Fixed rect selection in HDRP
```
JonMUnity, line 16: Which function? 

```suggestion
- [PBLD-228] Updated function to access shader property types.

### Fixed

- [PBLD-224] Fixed rect selection in HDRP.
```

jonbro: this line
`if (mat.shader.GetPropertyType(i) != UnityEngine.Rendering.ShaderPropertyType.Texture)`

its actually the `ShaderUtil.ShaderPropertyType.TexEnv` enum that was marked obsolete. This isn't as bad as I initially thought, it wouldn't have broken anything until the enum was removed, however I mistakenly clicked yes on the API upgrade dialog and this cause the error. Seems worth fixing now though, since the api that I switched it to use is available on all of the 6000 versions of unity.

saranyacodes (approved):
Looks good to me!

zeroyao (approved):

PR #613, opened 2025-07-08
Fixes for obsolete API warnings in Unity 6.2.

TimAidleyAtUnity:
### Purpose of this PR

This fixes [PBLD-245](https://jira.unity3d.com/browse/PBLD-245), where loading ProBuilder in to Unity 6.2 or later would result in a lot of warnings related to obsolete API.

The replacements for the obsolete TreeView API and EditorUtility.InstanceIDToObject do not exist prior to Unity 6.2, so in those cases I added pragmas to suppress the warnings. In the case of ShaderUtil.GetProperty___ I replaced the obsolete API calls with the correct ones as they exist as far back as Unity 2021.


zatzuri-unity (approved):
Tested on blank projects with 6.2 and 6.3. No obsolete API warnings pop up, looks good!

PR #615, opened 2025-07-15
Bugfix/pbld 226 hidden face selection fix

varinotmUnity:
### Purpose of this PR

This PR resolves a bug where single-click face picking on ProBuilder meshes failed if they were obscured by another GameObject, even when "Select Hidden" was enabled. The previous FaceRaycast implementation's internal next variable and looping logic caused it to often only consider the very first GameObject returned by the scene picker for a standard single click. This was problematic because if that initial GameObject was not a ProBuilder mesh, the ProBuilder picking routine would prematurely exit, preventing selection of the obscured ProBuilder face. The corrected approach now first collects all intersecting ProBuilder meshes and their hit faces under the mouse cursor. These ProBuilder hits are then internally sorted by distance, and the deep-click cycling logic is correctly applied exclusively to this filtered list of ProBuilder elements, ensuring reliable selection and improved interaction with obscured geometry.

### Links

[**Jira:**](https://jira.unity3d.com/browse/PBLD-226)

### Comments to Reviewers

Added unit tests

### DIFF
--- a/Editor/EditorCore/EditorSceneViewPicker.cs
+++ b/Editor/EditorCore/EditorSceneViewPicker.cs
@@ -406,6 +406,8 @@ namespace UnityEditor.ProBuilder
             return FaceRaycast(mousePosition, pickerOptions, allowUnselected, selection, 0, true);
         }
 
+        static List<(ProBuilderMesh mesh, Face face, float dist, int hash)> s_PbHits = new List<(ProBuilderMesh, Face, float, int)>();
+
         static float FaceRaycast(Vector3 mousePosition,
             ScenePickerPreferences pickerOptions,
             bool allowUnselected,
@@ -413,11 +415,12 @@ namespace UnityEditor.ProBuilder
             int deepClickOffset = 0,
             bool isPreview = true)
         {
-            GameObject pickedGo = null;
-            ProBuilderMesh pickedPb = null;
-            Face pickedFace = null;
+            GameObject candidateGo = null;
+            ProBuilderMesh candidatePb = null;
+            Face        candidateFace = null;
+            float       candidateDistance = Mathf.Infinity;
 
-            int newHash = 0;
+            s_PbHits.Clear();
 
             // If any event modifiers are engaged don't cycle the deep click
             EventModifiers em = Event.current.modifiers;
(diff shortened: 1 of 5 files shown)


Editor/EditorCore/EditorSceneViewPicker.cs
```diff
@@ -437,78 +442,121 @@ static float FaceRaycast(Vector3 mousePosition,
...
+                candidateFace = selectedHit.face;
+                candidateDistance = selectedHit.dist;
 
-                    // a prior hash was matched, this is the next. if
-                    // it's just the first iteration don't break (but do
-                    // set the default).
-                    if (next != 0)
-                        break;
+                // Update the newHash for the element actually picked after cycling/filtering
+                int newHash = selectedHit.hash;
```
TimAidleyAtUnity, line 511: nitpick: Couldn't you just combine this line with line 516?

Tests/Editor/Picking/FacePickerTests.cs
```diff
@@ -0,0 +1,383 @@
...
+    private ProBuilderMesh m_Mesh;
+    private Camera m_Camera;
+    private ScenePickerPreferences m_PickerPreferences;
+
+    // Helper to calculate the world-space center of a face
+    private Vector3 GetFaceCenterWorld(ProBuilderMesh mesh, Face face)
+    {
+        Vector3 centerLocal = Vector3.zero;
+        if (face.indexesInternal == null || face.indexesInternal.Length == 0)
+            return Vector3.zero; // Should not happen for valid ProBuilder faces
```
TimAidleyAtUnity, line 24: If this should never happen, is it worth making it into an assert instead?

TimAidleyAtUnity (changes requested):
I really like this PR. Very clear, and the comments are useful.

I made a couple of minor points about switching some code to asserts, and I suspect we should clear s_pbHits after use, but other than that, great job!

Tests/Editor/Picking/FacePickerTests.cs
varinotmUnity: I will use Assume instead, as this is not what we want to test

zatzuri-unity (approved):
- Tested repro steps
- Tested on multiple PB objects/faces with hidden and non hidden faces
- Tested with different angles
Looks good!

Tests/Editor/Picking/FacePickerTests.cs
TimAidleyAtUnity: What happens when an Assume fails? Will we get a notification?

TimAidleyAtUnity (approved):
Nice!

Tests/Editor/Picking/FacePickerTests.cs
varinotmUnity: https://docs.nunit.org/articles/nunit/writing-tests/Assumptions.html
The test will return as inconclusive

PR #616, opened 2025-08-01
Fixed a bug with UV autostitching where the position offset would not take into account the face rotation center offset.

varinotmUnity:
### Purpose of this PR

The calculation of the auto stitching was based on the offset between the destination UV center, and the source UV center. This offset is then applied to a face UV. 

The problem with the old calculation, was that when a rotation was needed for a target face, the bounds center could be slightly offseted, when the face was not a rectangle or cube (eg. a trapezoid), since the bounds are calculated using the min,max x and y. This slight offset would result in a misposition of the target face relative to the anchor, which would visually result in an offsetted texture. 

This was solved by making sure we take into account the rotation of the face when calculating the offset.

In the image below, we see that the rotation of the trapezoid modifies its center bound, which is not taken into the account in the calculation
image

The image below shows the old behaviour. notice the existing offset between the square and the trapezoid. It is the same offset as the image above:
image

After. We now take into account the rotation

autostiching

### Links

https://jira.unity3d.com/browse/PBLD-164

### Comments to Reviewers

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - [PBLD-242] Fixed a bug where edges were being incorrectly selected when one or both vertices were behind the camera's near plane, causing flipped lines and inconsistent selection behavior.
 - [PBLD-226] Fixed a bug where ProBuilder faces could not selected when obscured by another GameObject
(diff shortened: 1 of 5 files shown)


TimAidleyAtUnity (approved):
Congratulations on fixing this!

Also, I really like the comprehensive PR description you did, and the clear and well-commented tests you added.

zatzuri-unity (approved):

PR #619, opened 2025-08-13
fix issue where probuilder with quad topology instead of triangle would trigger out of bounds exception when being exported

varinotmUnity:
### Purpose of this PR

The degenerate triangle removal code introduced in [v6.0.6](https://github.com/Unity-Technologies/com.unity.probuilder/pull/606) assumed all submeshes used triangular topology, 
but ProBuilder meshes can contain mixed topologies (triangles, quads, etc.). The code incremented through indices 
by 3 and accessed `indexes[tri + 2]`, causing IndexOutOfRangeException when processing non-triangular submeshes 
like quads with indices arrays not divisible by 3. This specifically affected shapes like Door that use quad topology.

**Fix**: Added a topology check `if (submeshes[i].m_Topology == MeshTopology.Triangles)` before applying the 
degenerate triangle removal logic. This ensures the triangle-specific degenerate detection only runs on actual 
triangular submeshes, preventing the index bounds violation on quad/other topologies.

**Preserves Original Fix**: The degenerate triangle removal functionality remains completely intact for triangular 
submeshes - it still detects and removes zero-area triangles using cross-product magnitude checks. Non-triangular 
submeshes bypass this processing entirely, which is appropriate since the concept of "degenerate triangles" doesn't 
apply to quads or other topologies. The fix maintains backward compatibility while eliminating the crash.

**Result**: Door shapes and other mixed-topology meshes can now be exported without IndexOutOfRangeException, 
while triangular meshes continue to benefit from degenerate triangle cleanup and improved rendering quality.

### Links

https://jira.unity3d.com/browse/PBLD-251

### Comments to Reviewers

I did not check for degenerate quads in this PR, because I could not repro the issue the user was having with degenerate triangles (making the bloom flash weirdly)

### DIFF
--- a/Runtime/Core/ProBuilderMeshFunction.cs
+++ b/Runtime/Core/ProBuilderMeshFunction.cs
@@ -409,27 +409,40 @@ namespace UnityEngine.ProBuilder
                 // remove any indices that contain degenerate triangles
                 int numBadIndices = 0;
                 int[] indexes = submeshes[i].m_Indexes;
-                for (var tri = 0; tri < indexes.Length; tri += 3)
+
+                if (submeshes[i].m_Topology == MeshTopology.Triangles)
                 {
-                    Vector3 ab = positions[indexes[tri + 1]] - positions[indexes[tri]];
-                    Vector3 ac = positions[indexes[tri + 2]] - positions[indexes[tri]];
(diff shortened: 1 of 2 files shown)


zatzuri-unity (approved):
Tested the repro steps for this issue as well as the original repro steps from the degenerate triangles issues. I've also tested exporting different meshes

PR #622, opened 2025-08-21
Fix Material Warnings When Importing ProBuilder

zatzuri-unity:
### Purpose of this PR

Warnings pop up about materials being altered when one imports the ProBuilder package in Unity 6000.2 and above. This PR serves to update those materials to fix the warnings.

### Links

**Jira:**

### Comments to Reviewers

Tested by importing the local copy of ProBuilder with this fix on the latest versions of Unity 6000.2 and 6000.0


TimAidleyAtUnity:
Is this version of ProBuilder supported on Unity 2022.3?

Also, this doesn't get to the root of my the materials are being altered.

zatzuri-unity:
> Is this version of ProBuilder supported on Unity 2022.3?

2022.3 uses 5.2.x while Unity 6000.0 and above uses ProBuilder 6.x.x.

PR #623, opened 2025-08-25
Update CHANGELOG for version 6.0.7 release

zatzuri-unity:
### Purpose of this PR

Update CHANGELOG for version 6.0.7 release

### Links

**Jira:** n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [6.0.7] - 2025-08-29
```
TimAidley, line 8: We have the 29th off. Should this be the 28th, or does it not really matter?

zatzuri-unity: Right... I'll just change it so we can be consistent with what's stated on the release schedule, which happens to be the 28

PR #626, opened 2025-10-06
quick find and replace for InstanceID to EntityId

varinotmUnity:
### Purpose of this PR

The method GetInstanceID()  will be made obsolete in the very near future. We had to modify the code to not use GetInstanceID() , but  GetEntityId()  instead. https://unity.slack.com/archives/C7HJDB0A0/p1759330037063929

### Links

**Jira:**

### DIFF
--- a/Documentation~/workflows.md
+++ b/Documentation~/workflows.md
@@ -1,2 +1,0 @@
-# ProBuilder workflows
-


TimAidleyAtUnity (changes requested):
We need to make sure the code works fine for Unities 6.0 to 6.3.

PR #629, opened 2025-10-12
Making some Types and Properties public to allow building of custom tools with it

CortiWins:
### Purpose of this PR
The changes make three types public 
- PositionToolContext
- TextureToolContext
- CutTool

This allows to build shortcuts that probuilder into the respective tools and edit mode.

The changes one property public 
- MeshSelection.totalVertexCountOptimized

it is calculated and updated in the same way all other public statistics are, so making it public allows to build UI that shows statistics.


Editor/EditorCore/CutTool.cs
```diff
@@ -19,7 +19,7 @@
 namespace UnityEditor.ProBuilder
 {
     [EditorTool("Cut Tool", typeof(ProBuilderMesh), typeof(PositionToolContext))]
-    class CutTool : EditorTool
+    public class CutTool : EditorTool
```
lopezt-unity, line 22: We should make that `sealed`, so that it's possible to reference it but that should be all.
```suggestion
    public sealed class CutTool : EditorTool
```

Editor/EditorCore/ProBuilderToolsContexts.cs
```diff
@@ -18,7 +18,7 @@ static class ProBuilderToolManager
 
     [Icon("Packages/com.unity.probuilder/Content/Icons/EditableMesh/EditMeshContext.png")]
     [EditorToolContext("ProBuilder", typeof(ProBuilderMesh))]
-    class PositionToolContext : EditorToolContext
+    public class PositionToolContext : EditorToolContext
```
lopezt-unity, line 21: We should make that sealed, so that it's possible to reference it but that should be all.
```suggestion
    public sealed class PositionToolContext : EditorToolContext
```

Editor/EditorCore/ProBuilderToolsContexts.cs
```diff
@@ -407,7 +407,7 @@ static void SetColliderAction(MenuCommand command)
         }
     }
 
-    class TextureToolContext : EditorToolContext
+    public class TextureToolContext : EditorToolContext
```
lopezt-unity, line 410: We should make that sealed, so that it's possible to reference it but that should be all.
```suggestion
    public sealed class TextureToolContext : EditorToolContext
```

Editor/EditorCore/TextureMoveTool.cs
```diff
@@ -48,22 +47,25 @@ protected override void DoToolGUI()
...
 
             Handles.color = Color.green;
 
-            m_Position = Handles.Slider(m_Position, Vector3.up);
+            m_Position = Handles.Slider(m_Position, Vector3.up, HandleUtility.GetHandleSize(m_Position), Handles.ArrowHandleCap, SnapFactor);
 
             Handles.color = Color.red;
 
-            m_Position = Handles.Slider(m_Position, Vector3.right);
+            m_Position = Handles.Slider(m_Position, Vector3.right, HandleUtility.GetHandleSize(m_Position), Handles.ArrowHandleCap, SnapFactor);
```
lopezt-unity, line 68: These can be removed as I'm landing a fix on another PR :D

lopezt-unity (changes requested):

Editor/EditorCore/CutTool.cs
CortiWins: I'm a fan of sealing things that are not made to be extended. So thumbs up for that.

lopezt-unity:
Hey @CortiWins,
I will have to spend a little more time on this.
Making these class public is also exposing API from inside these class that does not need to be exposed 😓 
I will have to take a look at that and make some additional changes.

CortiWins:
@lopezt-unity I am not that surprised by the answer.

The whole concept of tools and contexts is a bit...semi public right now. I helped on an addon called "ProBuilderPlus" by ex-Unity ProBuilder creator GabrielW and that required a bit of reflection to use the ToolManager.Tool and ToolManager.Context functions.
Setting them via Types and reading and comparing the Type works fine as soon as yo have them.
But the ones Unity uses for different editor functions, including ProBuilder, are not public so it's having a public API, but lacking the access to really use it.

Maybe a compromise is a property that returns the type like 

```
public static ProBuilderContexts
{
  System.Type CutToolType => typeof(CutTool);
}
```
Because the use of knowing the exact Types is not changing something about their instances, but using them for setting and comparing ToolManager.Tool and ToolManager.Context throught  ToolManager.SetActiveTool(Type)/ToolManager.SetActiveContextType).

lopezt-unity:
> @lopezt-unity I am not that surprised by the answer.
> 
> The whole concept of tools and contexts is a bit...semi public right now. I helped on an addon called "ProBuilderPlus" by ex-Unity ProBuilder creator GabrielW and that required a bit of reflection to use the ToolManager.Tool and ToolManager.Context functions. Setting them via Types and reading and comparing the Type works fine as soon as yo have them. But the ones Unity uses for different editor functions, including ProBuilder, are not public so it's having a public API, but lacking the access to really use it.
> 
> Maybe a compromise is a property that returns the type like
> 
> ```
> public static ProBuilderContexts
> {
>   System.Type CutToolType => typeof(CutTool);
> }
> ```
> 
> Because the use of knowing the exact Types is not changing something about their instances, but using them for setting and comparing ToolManager.Tool and ToolManager.Context throught ToolManager.SetActiveTool(Type)/ToolManager.SetActiveContextType).

Sorry, I should express a little better :)
I totally agree that these class should be public :) but it should be done the right way and that might involve to have a look inside these classes and be certain that the correct APIs are exposed/internal :) 
It's just some extra work we'll have to do :) I need to plan for some extra time on that, but it'll be done :D

PR #660, opened 2026-03-13
[UUM-133530] Ensure "Set Double Sided" is enabled when active selection is proper

thomas-tu:

### Purpose of this PR

Implementation of the menu action was missing two elements:
- call to `base.enabled` that does some extra checks to validate that the selection context is correct with the provided `validSelectModes`. It's a common pattern in ProBuilder (look at all PB's menu action).
- `hasFileMenuEntry` needs to be set to false. Base implementation returns `true` but that would indicate that the custom action exists somewhere in the top menu as the code would try to resolve the action by looking up at the one from the top menu.

Also fixed the contextual menu builder so it pulls the status of the MenuAction. In cases where the action doens't have a fileMenuEntry, nor options, it was assumed that the option is always enabeld. I have added some tests to cover these cases.

### Links

**Jira:** UUM-133530


varinotmUnity (approved):
nice, users have been asking for this for a while on forums :)

unity-kristinn (approved):
suggested by u-pr test plan

> Face Selection Validation: Enter Face mode. With no faces selected, verify the action is disabled. Select one or more faces and verify the action becomes enabled.this seems to not be the case for me.

since it’s only a sample, i don’t think it matters much. is this something you would like to fix?

approving anyhow

CHANGELOG.md
```diff
@@ -10,7 +10,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - [UUM-133861] Fixed "Look rotation viewing vector is zero" log being spammed when holding shift while using a create tool such as Create Sprite.
+<<<<<<< bugfix/sample-menuaction-disabled
+- [UUM-133530] Fixed the `Set Double Sided` custom action in the Editor Sample, which was previously remaining disabled.
+- [UUM-133530] Ensured that the context menu respects the value of `MenuAction.enabled`.
+=======
 - [UUM-133531] Fixed component icons in Light theme.
+>>>>>>> master
```
varinotmUnity: your changelog has merging issues

PR #664, opened 2026-04-02
[UUM-138957] Remove runtime test dependency on UnityEditor APIs

lopezt-unity:
### Purpose of this PR

The PR changes the way Framework and Runtime tests are working in ProBuilder.
 
 Previously these supposedly runtime tests where using UnityEditor APIs, which was failing on some trunk tests.
 
 For more context:
 _Two probuilder test assembly definitions (Unity.ProBuilder.Tests.Framework.asmdef and Unity.ProBuilder.Tests.asmdef) had "includePlatforms": [] (all platforms) despite using UnityEditor.* APIs (AssetDatabase, EditorSceneManager, PackageManager.PackageInfo). These editor-only APIs are unavailable to runtime assemblies, causing CS0234/CS0103 compilation errors, that's why I wanted to merge those changes in my PR. By changing both "includePlatforms": ["Editor"] to match their actual dependencies, consistent with the existing Unity.ProBuilder.Editor.Tests.asmdef , the jobs no longer fail._

However, runtime tests should not be relying on Editor APIs. 
This PR update the runtime tests to get rid of the different Editor APIs used in it.

Most changes are:
- Runtime test asmdef does not rely on `Tests.Framework` anymore
- Needed methods from `TestUtility.cs` (in `Tests.Framework`) have been moved to `RuntimeUtility.cs`
- Assets used to compare a mesh edited in a test vs the expected result have been moved from the `Template` folder to a `Resources` one so that they can be loaded directly using `Resources.Load()`
- `RuntimeUtility.cs` have been updated accordingly

### Links

**Jira:** https://jira.unity3d.com/browse/UUM-138957

### Comments to Reviewers

All the tests have be re-run in the Editor (Editor and Playmode) and are all 🟢


xianenli30 (approved):
Risks reside in test behaviours with asset loading and test migration correctness. Approve with the condition of both editor and runtime tests passing and validated.

antoine-unity (approved):

PR #670, opened 2026-05-26
[UUM-133529] ProBuilder GameObjects not changing back to 3D shapes after Plane/Sprite transition

lopezt-unity:
#Purpose of this PR

  ##Problem

  When a ProBuilder GameObject (e.g., Arch, Cube) had its shape changed to Plane or Sprite, it could not be changed back to a 3D shape. The GameObject would remain flat instead of becoming a proper 3D mesh.

  ##Reproduction steps:
  1. Create any ProBuilder 3D GameObject (e.g., GameObject > ProBuilder > Arch)
  2. Change Shape to Plane or Sprite in the ProBuilder Shape component
  3. Change Shape back to any 3D shape
  4. Bug: GameObject remains flat instead of becoming 3D

  ##Root Cause

  The ProBuilderShape.SetShape() method had logic to flatten the Y dimension when transitioning TO Plane/Sprite shapes, but no corresponding logic to restore the Y dimension when transitioning FROM Plane/Sprite TO 3D shapes. This left m_Size.y at 0, causing the 3D shape to be built with zero height.

  ##Solution

  Added detection for 2D→3D shape transitions in SetShape(). When transitioning from a flat shape (Plane/Sprite) to a 3D shape, the Y dimension is now restored to 1.0 if it was  previously zero.

# Links

**Jira:** https://jira.unity3d.com/browse/UUM-133529

### Comments to Reviewers


modrimkus-unity (approved):
LGTM, though AI might have a point on Mathf.Epsilon. I'd check what has more precedent in the codebase (are we checking more against Mathf.Epsilon or 0.0001f for example).

PR #671, opened 2026-05-27
Bugfix/uum 141074 fix create shape gridsnap

lopezt-unity:
#Purpose of this PR

  ##Problem

  The Create Shape tool (Cube, Sphere, Plane, etc.) was ignoring grid snapping when hovering the cursor over the scene view before starting to create a shape. Grid snapping would only activate after the first mouse click (MouseDown event). This inconsistent behavior differed from the Create PolyShape tool, which correctly snapped to the grid during hover.

  Additionally, the Create PolyShape tool had gridSnapEnabled not set to true, preventing the tool from properly integrating with Unity's grid snapping system.

  ##Root Cause

  In ShapeState_InitShape.cs, the m_IsOnGrid flag was only calculated during the MouseDown event (when clicking to start shape creation), not during hover/mouse move events. Since DrawShapeTool.GetPoint() requires both m_IsOnGrid && EditorSnapSettings.gridSnapActive to be true for snapping to work, the  cursor position remained unsnapped during hover.

##Solutions

  ###Create Shape Tool Fix

  Refactored ShapeState_InitShape.DoState() to calculate the plane and grid snapping state for ALL mouse events (hover + click), not just on MouseDown:

  Key Changes:
  - Moved plane calculation and m_IsOnGrid determination outside the MouseDown-specific branch
  - Grid alignment check now happens during hover, before GetPoint() is called
  - On MouseDown, the plane is persisted to tool members for subsequent states
  - Eliminated code duplication by unifying hover and click logic

  Result: m_IsOnGrid is correctly set before GetPoint() is called during hover, enabling grid snapping throughout the shape creation workflow.

  ###Create PolyShape Tool Fix

  Added gridSnapEnabled override to PolyShapeTool class:
`  public override bool gridSnapEnabled => true;`

  This ensures the tool properly integrates with Unity's grid snapping system.

### Links

**Jira:** https://jira.unity3d.com/browse/UUM-141074

### Comments to Reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- [UUM-141074] Fixed an issue where the CreateShape and CreatePolyshape tools would not properly snap to the grid.
 - [UUM-133861] Fixed "Look rotation viewing vector is zero" log being spammed when holding shift while using a create tool such as Create Sprite.
 - [UUM-133859] Fixed an issue in URP projects where the Editor would recompile scripts when after a rectangle selection in ProBuilder. 
 - [UUM-133530] Fixed the `Set Double Sided` custom action in the Editor Sample, which was previously remaining disabled.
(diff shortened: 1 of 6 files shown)


modrimkus-unity (approved):
LGTM!

PR #672, opened 2026-05-28
[UUM-141406] fixing action overlay behavior

lopezt-unity:
##Purpose of this PR

  Jira: UUM-141406 (https://jira.unity3d.com/browse/UUM-141406)

  This PR fixes two related issues with ProBuilder action overlays:

  1. Overlay Display Issue: When an action overlay was already displayed in the SceneView and a new action was
  triggered, the overlay would not properly refresh or display. The fix adds a workaround that toggles displayed from
  false to true to ensure the overlay is properly added/refreshed in the view.
  2. Live Preview Toggle Not Updating Fields: When users toggled the "Live Preview" setting during an action, input
  fields (FloatField, IntegerField, Vector3Field) would not dynamically update their isDelayed property. Users had to
  restart the action for the setting to take effect.

  ###Solution:
  - Introduced a delayedPreviewChanged event in PreviewActionManager that fires when the live preview setting changes
  - Updated all affected geometry actions (BevelEdges, ExtrudeEdges, ExtrudeFaces, OffsetElements, SubdivideEdges,
  WeldVertices, GrowSelection) to subscribe to this event and update their field properties dynamically
  - Fixed the overlay display issue with a toggle workaround in ActionSettings.cs
  - Removed the previous inefficient approach of destroying and recreating the entire overlay UI

 ##Release Notes

  Fixed: UUM-141406 - Action overlays now display correctly when triggering a new action while an overlay is already shown in the SceneView. 

  ##Functional Testing status

  Manual Testing:
  - Tested all seven modified geometry actions (Bevel Edges, Extrude Edges, Extrude Faces, Offset Elements, Subdivide
  Edges, Weld Vertices, Grow Selection)
  - Verified toggling "Live Preview" immediately updates field behavior (delayed vs immediate input)
  - Verified starting actions when overlays are already displayed properly shows the new action overlay
  - Tested both delayed and immediate preview modes with each action
  - Confirmed field input behavior matches the live preview setting in all cases

  Existing Automation:
  ProBuilder has existing editor tests for action functionality. The overlay display and UI field behavior changes are
  UI-specific and covered by manual testing.

 ##Performance Testing Status

  Performance Impact: Minimal positive impact

  The change replaces an inefficient approach (destroying and recreating the entire overlay UI hierarchy) with an
  event-based approach that only updates the isDelayed property on existing fields. This reduces allocations and UI
  rebuilds when toggling the live preview setting.

  No performance testing required - this is a UI behavior fix with a minor efficiency improvement.

  ##Overall Product Risks

  Technical Risk: 1
  Isolated UI behavior changes in ProBuilder action overlays. The event-based approach is straightforward, and the
  overlay display workaround is localized. Low risk of regression.

  Halo Effect: 1
  Changes are isolated to ProBuilder's action overlay system and the seven modified geometry actions. Does not affect
  other ProBuilder functionality, other packages, or Unity Editor core systems.

### DIFF
--- a/Editor/MenuActions/Geometry/BevelEdges.cs
+++ b/Editor/MenuActions/Geometry/BevelEdges.cs
@@ -62,10 +62,15 @@ namespace UnityEditor.ProBuilder.Actions
                 }
             });
 
-            PreviewActionManager.delayedPreviewChanged += () =>
+            System.Action onDelayedPreviewChanged = () =>
             {
                 floatField.isDelayed = PreviewActionManager.delayedPreview;
             };
+            PreviewActionManager.delayedPreviewChanged += onDelayedPreviewChanged;
+            root.RegisterCallback<DetachFromPanelEvent>(evt =>
+            {
+                PreviewActionManager.delayedPreviewChanged -= onDelayedPreviewChanged;
+            });
             root.Add(floatField);
 
             return root;
--- a/Editor/MenuActions/Geometry/BevelEdges.cs
+++ b/Editor/MenuActions/Geometry/BevelEdges.cs
@@ -61,6 +61,11 @@ namespace UnityEditor.ProBuilder.Actions
                     PreviewActionManager.UpdatePreview();
                 }
             });
+
+            PreviewActionManager.delayedPreviewChanged += () =>
+            {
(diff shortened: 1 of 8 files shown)


modrimkus-unity (approved):
LGTM!

PR #676, opened 2026-06-02
[UUM-138261] Fix move handle exponentially moving the selected elements

lopezt-unity:
### Purpose of this PR

This PR fixes an issue with the Move handle, creating a accumulation of the move deltas when using the handles. This was resulting in an exponential motion of the selected element.

This was caused by a wrong test in the rotation handle type, causing a cache invalidation in the middle of the process. This PR fixes that and focuses on getting PivotRotation.Local and World from unity as they are the only valid ones. This prevent that the use of new pivot types (such as grid or custom) messes up with ProBuilder state.

This PR also added tests to validate that PB pivot types are aligned with these changes.

### Links

**Jira:** https://jira.unity3d.com/browse/UUM-138261

### Comments to Reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changes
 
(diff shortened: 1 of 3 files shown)


modrimkus-unity (approved):
LGTM

PR #686, opened 2026-06-10
Publication/fixing assets serialized versions

lopezt-unity:
**DO NOT FORGET TO INCLUDE A CHANGELOG ENTRY**

### Purpose of this PR

[Brief summary of the changes in this PR.]

### Links

**Jira:**

### Comments to Reviewers

[List known issues, planned work, provide any extra context for your code.]

PR #697, opened 2026-08-25
Fix ProBuilderDefault and Checker materials incorrectly referencing the PB vertex color shader

modrimkus-unity:
### Purpose of this PR

PR fixes a regression (from https://github.com/Unity-Technologies/com.unity.probuilder/pull/686) where, after updating PB to 6.1.1 or above, meshes with materials that use the `PB6/Standard Vertex Color` shader (including the built-in `ProBuilderDefault.mat` and `Checker.mat`) would render in magenta. 

This happened because the breaking PR, after migrating large number of meta files, did not refresh the built-in materials, causing the link between the materials and the `Standard Vertex Color.shadergraph` asset to break. The migrated shgraph asset's meta file dropped: 
```
  fileIDToRecycleName:
    4800000: MainAsset
```
But the materials continued to point to 4800000: 
```
 m_Shader: {fileID: 4800000, guid: 8547ec39534635c4289065e5c5033f43, type: 3}
 ```
 

Various actions would indirectly fix the issue, for example, creating a new PB cube, due to it hitting codepath where the shared built-in material is assigned a newly loaded shader, fixing the broken reference, for example:

```
var material = (Material) Resources.Load("Materials/ProBuilderDefault", typeof(Material));
material.shader = Shader.Find("ProBuilder6/Standard Vertex Color");    <- overwrites the outdated reference
```

The fix updates the `ProBuilderDefault.mat` and `Checker.mat` assets to point to the correct fileID of the `PB6/Standard Vertex Color` shader.

### Links

**Jira:**. https://jira.unity3d.com/browse/UUM-150702

### Comments to Reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- [UUM-150702] Fixed the ProBuilderDefault and Checker materials referencing a stale shader ID, causing ProBuilder meshes to render magenta when a scene was opened.
 - [UUM-148237] Fixed the "Lightmap UVs Settings" foldout in ProBuilder Preferences not opening when clicking its title.
 - [UUM-133528] Fixed an issue where using some UV Editor actions would clear a mesh's Lightmap UVs without regenerating them.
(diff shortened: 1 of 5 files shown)


antoine-unity (approved):
LGTM

PR #698, opened 2026-08-26
Bring in trunk's CoreCLR compatibility changes (shadow-package sync)

lopezt-unity:
## What this does

Trunk made some changes to keep ProBuilder working with Unity's new CoreCLR editor, and those changes got mirrored into this repo automatically by the shadow-package bot (see PR #695). This PR brings those changes into `master`, since that bot PR landed on `release/6.1` instead, which isn't where we actually do ongoing development anymore.

Two kinds of changes came from trunk:

1. **Shader fix**: 16 shader files had their fragment output changed from the old `COLOR` semantic to the modern `SV_Target` semantic. Purely a naming/compatibility fix, no visual or behavioral change.
2. **Disabling some tests under CoreCLR**: 8 test files got a new attribute (`UnityCoreClrExplicitDisabled`) added to tests that currently fail when Unity runs on CoreCLR instead of Mono. This doesn't fix the underlying issues, it just stops those specific tests from running in CoreCLR test passes until someone fixes the real problem (tracked in UUM-148933 and UUM-148935).

## What I changed on top of the raw trunk changes

The attribute used to disable those tests (`UnityCoreClrExplicitDisabled`) is brand new and only exists starting in Unity 6000.7. This package still supports Unity all the way back to 6000.0, so using that attribute unconditionally would have broken compilation of our whole test suite on any older, currently-supported Unity version.

I wrapped every use of that attribute in a `#if UNITY_6000_7_OR_NEWER` check, so:
- On Unity 6000.7 and newer, it behaves exactly as trunk intended (skips those tests under CoreCLR).
- On older Unity versions, the attribute is simply left out. That's fine, since CoreCLR isn't even an option on those older versions, so there's nothing to disable there.

## Why master and not release/6.1

The shadow-package bot's PR (#695) targets `release/6.1`, but that branch diverged from `master` a while back and isn't where new package work lands. Posted a question in #devs-pets about whether the bot's sync target should change, or if this manual "catch it and move it to master" step is expected going forward.

## Testing

No functional behavior changes for supported use cases — this only affects shader semantics (cosmetic naming fix) and test execution under CoreCLR specifically. Recommend a normal CI run to confirm the test assembly still compiles cleanly on the oldest supported Editor version as well as on 6000.7+.

### DIFF
--- a/Content/Shader/EdgePicker.shader
+++ b/Content/Shader/EdgePicker.shader
@@ -49,7 +49,7 @@ CGPROGRAM
                 return o;
             }
 
-            float4 frag (v2f i) : COLOR
+            float4 frag (v2f i) : SV_Target
             {
                 return i.color;
             }
--- a/Content/Shader/FaceHighlight.shader
+++ b/Content/Shader/FaceHighlight.shader
@@ -45,7 +45,7 @@ Shader "Hidden/ProBuilder/FaceHighlight"
                 return o;
             }
 
-            half4 frag (v2f i) : COLOR
+            half4 frag (v2f i) : SV_Target
             {
                 i.pos.xy = floor(i.pos.xy * 1) * .5;
                 float checker = -frac(i.pos.x + i.pos.y);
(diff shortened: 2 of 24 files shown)


modrimkus-unity (approved):
LGTM. Thanks for the succinct explanation on what these changes are and why.

PR #700, opened 2026-09-11
[UUM-148935] Make VectorHash quantization runtime-independent

lopezt-unity:
## Purpose of this PR

Ports the fix from [unity/unity PR 123982](https://github.cds.internal.unity3d.com/unity/unity/pull/123982), which fixed [UUM-148935](https://jira.unity3d.com/browse/UUM-148935) and [UUM-133532](https://jira.unity3d.com/browse/UUM-133532) (formerly UUM-111993) in the vendored `ShadowPackages/com.unity.probuilder` snapshot in the engine repo. That PR's own description notes the fix must land here too, or it will be silently reverted on the next ShadowPackages refresh.

`VectorHash.HashFloat` hashed a float via `(ulong)(f * FltCompareResolution)`. Casting an out-of-range or NaN float to `ulong` is unspecified by ECMA-335: Mono on x64 wraps around, CoreCLR (and Mono on Arm64) saturate — collapsing every negative product to 0. That single cause explains all three `IntVectorTests` failures. `HashFloat` is replaced by `VectorHash.RoundToInt`, which maps NaN to 0, clamps explicitly instead of relying on the cast, and rounds half to even — deterministic on every runtime and architecture.

### Second defect, present on every runtime today

`IntVec2/IntVec3/IntVec4.round` quantized via `Convert.ToInt32(v * FltCompareResolution)` while `HashFloat` truncated — a `GetHashCode`/`Equals` contract violation. `SharedVertex.GetSharedVerticesWithPositions` and `ObjExporter` use `IntVec3` as a `Dictionary` key, so coincident vertices could land in different buckets and silently fail to weld. Routing both through `RoundToInt` fixes this too, and drops an `OverflowException` that `round()` threw on NaN/infinite positions.

### Test changes

- The three `IntVectorTests` no longer need a CoreCLR-disable guard, since the underlying hashing bug — not CoreCLR itself — was the cause.
- `TestVectorHashOverflow`'s expectations are updated to the new well-defined values (and a `NegativeInfinity` case is added).
- Added `TestEqualIntVec3SharesHashCode`, which asserts the `Equals`/`GetHashCode` agreement directly.

## Release Notes

See CHANGELOG.md.

## Reviewed by

Reviewed against the merged engine-repo PR (unity/unity#123982) and the linked Jira ticket (UUM-148935); logic and test-expectation changes were independently re-derived and checked for consistency rather than copied blind.


modrimkus-unity (approved):
LGTM!

